PR #268, opened 2020-01-09
Update install.ps1

whitetiger45:
Add Powershell version check to handle different commands to prevent failing of BoxStarter install on Windows 10 VM's downloaded as of 1-9-2020.


war59312:
What is the deal with this?

Close or approve already?

htnhan:
FLARE-VM don't officially support windows 10 yet.

PR #311, opened 2020-07-17
Removed citrix-receiver package

mwilliams31:


htnhan:
Packages.csv fixed in PR332 (v2.3.4)

PR #392, opened 2022-01-21
Fix typo in README.md

royreznik:
Just a small change that caught my eye when I installed the vm :)


mr-tz (approved):

mr-tz:
Thanks for the suggestion. This is addressed as part of the upcoming new installer script.

Issue #415, opened 2022-08-24
Myget.org resolution errors


chris4a50:
Recently when attempting to build FLARE-VM I continuously hit the following error after the first reboot:

`The remote name could not be resolved: '[www.myget.org](https://www.myget.org/)'`

It seems that the www.* subdomain of myget.org has either been deprecated or modified; changing the www.myget.org to simply https://myget.org/* on lines 309 and 360 resulted in a successful run of the installer. 

I'm not sure if this is a permanent or temporary change on the part of MyGet, but it works for now!

vm-packages:
Thank you for your feedback! We've been working on major updates to FLARE VM over the last year.
The now revamped FLARE VM has just been released and will make the project more open and maintainable.
Please check out our blog post at https://www.mandiant.com/resources/blog/flarevm-open-to-public and give the new installation a try.

If this problem still persists with the new installation, please report:
- new tools or tool-related issues at https://github.com/mandiant/VM-Packages/issues
- ideas and issues related to the installer script and configuration at https://github.com/mandiant/flare-vm/issues

Please note that we use this message to close all legacy issues in this repository. We look forward to your feedback and support for the next generation of FLARE VM.

PR #416, opened 2022-08-24
Fix myget resolution issue

chris4a50:
nix the www.* for myget's domain, resolves/closes https://github.com/mandiant/flare-vm/issues/415


mr-tz:
Thanks for the improvement. We've incorporated this into the upcoming new installer script.

PR #419, opened 2022-11-02
Disable automatic Microsoft Store updates to UWP apps.

BretMcDanel:
Disable automatic Microsoft Store updates to UWP apps.


mr-tz:
Thanks for the suggestion. We've added this to the upcoming new installer script.

PR #438, opened 2023-02-06
Fix not detecting stock VMWare/Hyper-V installations.

shavitush:
While installing on my VM (VMWare Workstation 17 Pro, Windows 10 Pro 64-bit 22H2), I've noticed how the installer was thinking my VM is hardened.

image

In my case, the computer system model was `"VMWare20,1"` which wouldn't pass the string equation of `== "VMware Virtual Platform"` of course.

This PR fixes the bug by checking if the computer system model contains any of the specified models, as well as changing the string of VMware's model to `"VMware"` to ensure success in all cases.

In addition, I've also added Hyper-V's model.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -188,8 +188,18 @@ if (-not $noChecks.IsPresent) {
     }
 
     # Check if system is a virtual machine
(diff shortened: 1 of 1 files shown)


mr-tz (approved):
thanks, this looks good to me

PR #441, opened 2023-03-09
updated README to include a note about setting execution policy for current scope

joshterrill:
When I was going through the README steps, I ran into an issue when setting the execution policy, so I added a note in the steps on how to fix it if other users run into the same issue.


mr-tz (approved):
thank you

PR #443, opened 2023-03-23
Fix Tamper Protection checking section

Hrishikesh7665:
Added **try-catch** block and **-ErrorAction** Stop parameter around **Get-ItemProperty** cmdlet to handle errors during execution. Catch block displays message indicating Tamper Protection is not enabled or detected, providing user with more information about potential issues during installation process.


mr-tz:
thanks!

Ana06 (approved):
Thanks! 👍

PR #469, opened 2023-06-28
config: remove pestudio

### LINKED ISSUE #468: pestudio breaks too often
We are downloading pestudio from the link https://www.winitor.com/tools/pestudio/current/pestudio.zip (which doesn't include the version) in the page https://www.winitor.com/download2 This tools gets often updated, which is a good thing as it is actively maintained, but breaks our package. I have asked the maintainer about using a link that includes the version and keep the old version for some time to avoid that every release breaks the package: https://twitter.com/ochsenmeier/status/1665675755244736514
Screenshot 2023-06-06 at 11 22 21

In the meanwhile the package has broken again, so I propose we remove it from the default flare-vm configuration to avoid making the installation fail, which is getting users confused. I hope we can re-add it soon.

Ana06:
pestudio uses a url that doesn't include the version and the package breaks with every update as the hash changes. As the tool is updated often, this packages fails to install often and confuses users.

Closes https://github.com/mandiant/flare-vm/issues/468

### DIFF
--- a/config.xml
+++ b/config.xml
@@ -41,7 +41,6 @@
         <package name="pebear.vm"/>
         <package name="peid.vm"/>
         <package name="pesieve.vm"/>
-        <package name="pestudio.vm"/>
         <package name="processdump.vm"/>
         <package name="regshot.vm"/>
         <package name="rundotnetdll.vm"/>


mr-tz (approved):
Huh, Ana slinging the hammer 😄 
Hope they can provide a consistent link soon and then we can re-add.

Ana06:
I hope we can re-add pestudio as default soon as well. It is still possible to install it using `choco install pestudio --ignore-checksums`

PR #470, opened 2023-06-29
Add default packages

Ana06:
Add pestudio and other packages recently added to [VM-Packages](https://github.com/mandiant/VM-Packages). 😄 

Regarding pestudio, see https://github.com/mandiant/VM-Packages/pull/471, thanks to marc ochsenmeier for providing the new URL that prevents the package to fail so that we can re-add the package. 💘 

@mr-tz am I missing any other package that should be in the default flare-vm installation? 🤔

### DIFF
--- a/config.xml
+++ b/config.xml
@@ -41,6 +41,7 @@
         <package name="pebear.vm"/>
         <package name="peid.vm"/>
         <package name="pesieve.vm"/>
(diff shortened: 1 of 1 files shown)


mr-tz:
Looks good, let's set a regular reminder to revisit the default config list!

mr-tz (approved):

PR #485, opened 2023-08-24
Adding check for usernames with spaces 

HuskyHacks:
Adding a small check in the install script that will detect spaces in usernames and exit gracefully if one is identified:

Detecting a space and exiting:
2023-08-24 13_48_16-go-101-win - VMware Workstation

...and continuing with the installation if the username looks good:
2023-08-24 13_45_51-go-101-win - VMware Workstation

... and this check is encapsulated in the block of the rest of the checks, so the `-noChecks` param will bypass it, even if the username is non-compliant:
image

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -226,6 +226,19 @@ if (-not $noChecks.IsPresent) {
     if ($response -notin @("y","Y")) {
         exit 1
     }
+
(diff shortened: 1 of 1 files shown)


mr-tz (approved):
awesome, thanks!

PR #488, opened 2023-09-19
Update README.md

wand3rlust:
Discovered this link through HuskyHacks's discord while searching for a clean way to disable Defender. It solves the issue of Defender magically enabling itself and can be easily enabled when needed with just a few tweaks.

Tested on Windows 10 Enterprise Eval 22H2 (19045.3324) and it works, been 10+ days without Defender popping up.


mr-tz (approved):
lgtm, thanks!

PR #504, opened 2023-10-10
Added check for Internet connectivity

wand3rlust:
Added check to detect Internet connectivity (via ICMP) using 'Test-Connection' cmdlet. The response is a boolean value and if said value is 'false', installer exits.

**[Tested on]:**

1. **OS:** Windows 10 19045.3324
2. **PowerShell Version:** 5.1.19041.3031

**[Case1]: Internet connectivity detected**

**[Case 2]: Internet connectivity not detected**

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -111,6 +111,7 @@ if (-not $noChecks.IsPresent) {
             Write-Host "`t[+] Hint: https://support.microsoft.com/en-us/windows/prevent-changes-to-security-settings-with-tamper-protection-31d51aaa-645d-408e-6ce7-8d7f8e593f87" -ForegroundColor Yellow
             Write-Host "`t[+] Hint: https://www.tenforums.com/tutorials/123792-turn-off-tamper-protection-windows-defender-antivirus.html" -ForegroundColor Yellow
             Write-Host "`t[+] Hint: https://github.com/jeremybeaume/tools/blob/master/disable-defender.ps1" -ForegroundColor Yellow
+            Write-Host "`t[+] Hint: https://lazyadmin.nl/win-11/turn-off-windows-defender-windows-11-permanently/" -ForegroundColor Yellow
             Write-Host "`t[+] You are welcome to continue, but may experience errors downloading or installing packages" -ForegroundColor Yellow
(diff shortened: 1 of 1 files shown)


install.ps1
```diff
@@ -241,6 +242,21 @@ if (-not $noChecks.IsPresent) {
     }
 }
 
+# Check Internet connectivity and return boolean value, exit if value is 'false'
```
Ana06, line 245: The internet check needs to be inside the `if (-not $noChecks.IsPresent)`.

Ana06 (changes requested):
@wand3rlust this is a great addition!

I am planing to add some other checks and improvements to the installer in the next few days (Powershell version check, update to use the new debloat and installer packages, etc.), that will cause merge conflicts with this PR. It would be good if you could address the changes requested quick, so that we can merge this PR before the rest 😉

install.ps1
```diff
@@ -241,6 +242,21 @@ if (-not $noChecks.IsPresent) {
...
+Write-Host "[+] Checking for Internet connectivity..."
+$connectionStatus =  Test-Connection 1.1.1.1 -Quiet
```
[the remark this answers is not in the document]
Ana06: [nitpick] even better would be to check that the computer can reach myget or github as we rely on them in the installer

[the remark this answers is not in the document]
wand3rlust: Added check for `github.com`, tried with `myget.org` & `www.myget.org` however they both seem to be blocking the ICMP packets.

install.ps1
wand3rlust: Done

install.ps1
```diff
@@ -239,6 +240,22 @@ if (-not $noChecks.IsPresent) {
     } else {
         Write-Host "`t[+] Username '$extractedUsername' does not contain any spaces." -ForegroundColor Green
     }
+
+# Check Internet connectivity and return boolean value, exit if value is 'false'
```
Ana06, line 244: This should be indented. But I think it is better to add a linter for this: https://github.com/mandiant/flare-vm/issues/507

Ana06 (approved):
@wand3rlust thanks for the contribution and for addressing the changes that fast!

PR #490, opened 2023-09-28
Update FLARE-VM logo & default config

Ana06:
We got a new cool logo! 🥳  Update the logo in the README and in the background image of the VM. Add also recently created packages in VM-Packages to the default FLARE-VM configuration.

Closes https://github.com/mandiant/flare-vm/issues/483

PR #497, opened 2023-10-05
Update README.md

MalwareMechanic:
Fixed logo URL

PR #508, opened 2023-10-12
Fixed broken link of FLARE-VM Background

wand3rlust:
Fixed a broken link due to the recent changes in logo (https://github.com/mandiant/flare-vm/pull/490)

**Issue section:**
flare

P.S: The designer(s) deserve Kudos, new logo looks amazing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ Below are links for post-installation tweaks for Windows 10+.
 
 We do encourage you to download and set your background to the FLARE VM logo!
 <p align="center">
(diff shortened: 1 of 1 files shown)


Ana06 (approved):
Yes, https://github.com/mandiant/flare-vm/pull/497 tried to fix this already. As I said in https://github.com/mandiant/flare-vm/pull/497#discussion_r1348460555, I think we should update the documentation as this is probably not needed. But it likely makes sense to update the documentation after the use of the new installer (https://github.com/mandiant/flare-vm/issues/510) and it looks ugly to leave this broken till then (even if it is for few days). So let's merge this. Thanks @wand3rlust!!

PR #513, opened 2023-10-13
Use debloat.vm and installer.vm

### LINKED ISSUE #510: Update to use new installer and debloat packages
### Details

This will allow us to automate many of the manual steps for creating a VM. We may want to create extra issues for the steps that are more elaborated to automate.

Ana06:
Use new debloat and installer.vm packages that will allow us to automate many of the manual steps for creating a VM.

Closes https://github.com/mandiant/flare-vm/issues/510


install.ps1
```diff
@@ -845,18 +846,25 @@ refreshenv
 Write-Host "[+] Installing shared module..."
 choco install common.vm -y --force
 refreshenv
+
+# Use single config
```
Ana06, line 850: At least at this point, I think it is the easiest alternative (also for user modifying the config) to have a single config file. We can discuss this further after this has been merged (maybe also with commando-vm)

PR #514, opened 2023-10-13
Improve checks

Ana06:
Perform checks in a more logical order, including:
- Check first the Powershell version as some of the code may not work otherwise. Exit instead of using `-notin` that was introduced in Powershell 3 (and some user had previously used to try to install FLARE-VM).
- Ask about the snapshot after verifying the rest of the checks.

Update and document the Windows versions used currently for testing.
Screenshot 2023-10-13 at 15 59 35

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -78,6 +78,17 @@ $desktopPath = [Environment]::GetFolderPath("Desktop")
 Set-Location -Path $desktopPath -PassThru | Out-Null
 
 if (-not $noChecks.IsPresent) {
+    # Check PowerShell version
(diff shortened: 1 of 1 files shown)


install.ps1
```diff
@@ -78,6 +78,17 @@ $desktopPath = [Environment]::GetFolderPath("Desktop")
 Set-Location -Path $desktopPath -PassThru | Out-Null
 
 if (-not $noChecks.IsPresent) {
```
wand3rlust, line 80: Will it be better if the PowerShell version check is outside the skip parameter so it can't be skipped? I think that would at the very least lead to less chaos at later stages.

Issue #393, opened 2022-02-19
How to really disable Windows Defender on Win11


pieterhouwen:
The installer script still ran into errors disabling Defender even when following the steps in the guide that is linked in the README. 

The way I finally killed Windows Defender is by following "Using RegEdit to disable Windows Defender in Windows 11" on the site: https://lazyadmin.nl/win-11/turn-off-windows-defender-windows-11-permanently/

TL;DR:

```
An alternative way to disable Windows Defender in Windows 10 or 11 is to use the registry. For this, you will also need to boot into safe mode. Follow step 1 from the previous chapter to boot into safe mode.

Once booted in safe mode:

Press Windows key + R
Type regedit <enter> to open the registry
Navigate to HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Service
Change the following folders the key Start to 4
Sense
WdBoot
WdFilter
WdNisDrv
WdNisSvc
WinDefend

Once done restart back to normal boot
```

Hope this helps!

vm-packages:
Thank you for your feedback! We've been working on major updates to FLARE VM over the last year.
The now revamped FLARE VM has just been released and will make the project more open and maintainable.
Please check out our blog post at https://www.mandiant.com/resources/blog/flarevm-open-to-public and give the new installation a try.

If this problem still persists with the new installation, please report:
- new tools or tool-related issues at https://github.com/mandiant/VM-Packages/issues
- ideas and issues related to the installer script and configuration at https://github.com/mandiant/flare-vm/issues

Please note that we use this message to close all legacy issues in this repository. We look forward to your feedback and support for the next generation of FLARE VM.

PR #530, opened 2023-11-16
Updated Windows Defender Disable Methods in README.

0xdf223:
Added three working windows defender disable methods to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,10 +22,14 @@ The VM should satisfy the following requirements:
 * Prepare a Windows 10+ virtual machine
   * Install Windows in the virtual machine, for example using the raw Windows 10 ISO from https://www.microsoft.com/en-us/software-download/windows10ISO
   * Ensure the [requirements above](#requirements) are satisfied, including:
-    * Disable Windows Updates (at least until installation is finished)
-      * https://www.windowscentral.com/how-stop-updates-installing-automatically-windows-10
+    * Disable Windows Updates (at least until installation is finished). Windows is constantly trying to make it harder to do this. In Windows 10 and 11, there are a couple ways to achieve this:
+      * Booting into safe mode and taking ownership of the defender executable and locking all user out (as described [here](https://lazyadmin.nl/win-11/turn-off-windows-defender-windows-11-permanently/)).
+      * Booting into safe mode and disabling the defender-related services (as described [here](https://lazyadmin.nl/win-11/turn-off-windows-defender-windows-11-permanently/)).
+      * [This](https://github.com/AveYo/LeanAndMean/blob/main/ToggleDefender.ps1) powershell script from AveYo.
     * Disable Tamper Protection and any Anti-Malware solution (e.g., Windows Defender), preferably via Group Policy.
+      * The methods above for defender will also take care of Tamper Protection.
       * [https://stackoverflow.com/questions/62174426/how-to-permanently-disable-windows-defender-real-time-protection-with-gpo](https://superuser.com/a/1757341)
+      
 * Take a VM snapshot so you can always revert to a state before the FLARE-VM installation
 
 ### FLARE-VM installation
--- a/README.md
+++ b/README.md
@@ -22,14 +22,13 @@ The VM should satisfy the following requirements:
 * Prepare a Windows 10+ virtual machine
   * Install Windows in the virtual machine, for example using the raw Windows 10 ISO from https://www.microsoft.com/en-us/software-download/windows10ISO
   * Ensure the [requirements above](#requirements) are satisfied, including:
-    * Disable Windows Updates (at least until installation is finished). Windows is constantly trying to make it harder to do this. In Windows 10 and 11, there are a couple ways to achieve this:
+    * Disable Windows Updates (at least until installation is finished)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,10 +22,14 @@ The VM should satisfy the following requirements:
 * Prepare a Windows 10+ virtual machine
   * Install Windows in the virtual machine, for example using the raw Windows 10 ISO from https://www.microsoft.com/en-us/software-download/windows10ISO
   * Ensure the [requirements above](#requirements) are satisfied, including:
-    * Disable Windows Updates (at least until installation is finished)
-      * https://www.windowscentral.com/how-stop-updates-installing-automatically-windows-10
```
Ana06, line 26: what is the problem with these instructions?

mr-tz: They don't work anymore apparently - https://twitter.com/0xdf_/status/1725287455694163988

0xdf223: At least they didn't work for me. Using GPOs would require a reboot, and on reboot, it would be back and running again.

Ana06: @0xdf223 are you using Windows 10? Because I think I used that guide few weeks ago and it worked. 🤔

0xdf223: Yes, win 10. Are you fully updated before trying to do this?

Maybe the best option would be to leave the old one and add the new references? I'm not the first person to run into this issue: https://github.com/mandiant/flare-vm/issues/393

Ana06: > Are you fully updated before trying to do this?
No, I normally do not update my malware analysis VMs as vulnerability fixes may break malware functionality I may want to analyse.

I removed some links in https://github.com/mandiant/flare-vm/pull/518/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L28-L36 with the goal to simplify the documentation and leave a link I had tested to work (hoping this will work for other as well and be more straighforward to implement). But of course if this doesn't work, we should add extra documentation. Do you think we need all the links you are adding (thinking of making as easy and straighforward to users as possible).

mr-tz: I think it's fair to have fallback links - especially given how annoying Defender can be to disable reliably any tested/helpful resource is appreciated.

README.md
```diff
@@ -22,10 +22,14 @@ The VM should satisfy the following requirements:
...
+    * Disable Windows Updates (at least until installation is finished). Windows is constantly trying to make it harder to do this. In Windows 10 and 11, there are a couple ways to achieve this:
+      * Booting into safe mode and taking ownership of the defender executable and locking all user out (as described [here](https://lazyadmin.nl/win-11/turn-off-windows-defender-windows-11-permanently/)).
+      * Booting into safe mode and disabling the defender-related services (as described [here](https://lazyadmin.nl/win-11/turn-off-windows-defender-windows-11-permanently/)).
+      * [This](https://github.com/AveYo/LeanAndMean/blob/main/ToggleDefender.ps1) powershell script from AveYo.
     * Disable Tamper Protection and any Anti-Malware solution (e.g., Windows Defender), preferably via Group Policy.
+      * The methods above for defender will also take care of Tamper Protection.
       * [https://stackoverflow.com/questions/62174426/how-to-permanently-disable-windows-defender-real-time-protection-with-gpo](https://superuser.com/a/1757341)
+
```
Ana06, line 32: ok, let's add some more links then! What about doing it this way (as it seems some of them solve several items in the list)?

```suggestion
    * Disable Windows Updates (at least until installation is finished), Tamper Protection and any Anti-Malware solution (e.g., Windows Defender), preferably via Group Policy. Resources:
      * https://www.windowscentral.com/how-stop-updates-installing-automatically-windows-10
      * Boot into safe mode and take ownership of the defender executable and lock all user out as described in https://lazyadmin.nl/win-11/turn-off-windows-defender-windows-11-permanently.
      * Boot into safe mode and disable defender-related services as described in https://lazyadmin.nl/win-11/turn-off-windows-defender-windows-11-permanently.
      * Run https://github.com/AveYo/LeanAndMean/blob/main/ToggleDefender.ps1.
      * [https://stackoverflow.com/questions/62174426/how-to-permanently-disable-windows-defender-real-time-protection-with-gpo](https://superuser.com/a/1757341)
```

@0xdf223 are you ok with this change? If so, are you able to make this change and sign the CLA or should I make the change?

README.md
Ana06: @0xdf223 all the links you have added seems to be related to disabling Windows Defender, but you put them under _Disable Windows Updates_ (and the link you are removing is also related to this). Am I understanding something wrong?

chrisbensch:
So far, the best way I've found to remove Defender from both Win10 and Win11 uses Psexec, Sordum DefenderControl, and disable-defender.ps1 from https://github.com/jeremybeaume/tools/tree/master.  Here are my steps:  Open Virus settings and disable real-time monitoring and tamper protection, using psexec to run sordum defendercontrol to disable Defender, then again using psexec to run disable-defender.ps1.  It will finish after a reboot.  If you want to do it all in one fell swoop, you need to modify disable-defender.ps1 and manually specifiy the path to psexec.exe.  Works on the most recent version of 10/11 fully updated.

README.md
1AM3RR0R: a proper way to disable Defender on new/updated OS versions (including 11) is described [here](https://woshub.com/disable-windows-defender-antivirus/).

Ana06:
Closing in favor of https://github.com/mandiant/flare-vm/pull/597

PR #531, opened 2023-11-21
Add Custom Start Layout for taskbar pinning

### LINKED ISSUE #512: Configuration: Pin taskbar icons
### Details

Pin the following icons to the taskbar:
- File Explorer (if possible with the Tools folder)
- cmd.exe (Run as admin)
- Chrome
- IDA
- fakenet (Run as admin)
- CFFExplorer
- procexp (Run as admin)
- Process Monitor (Run as admin)
- notepad++

This is to be add to the default config file (config.xml). We can probably implement this similarly as in https://github.com/mandiant/VM-Packages/blob/6423a6184172bd7a2af51eb1554551d8f45d7553/packages/debloat.vm/tools/win10.xml#L164

Requires https://github.com/mandiant/flare-vm/issues/510

emtuls:
This adds the custom Start Layout xml to FlareVM which works in tandem with https://github.com/mandiant/VM-Packages/pull/748

This all together resolves https://github.com/mandiant/flare-vm/issues/512

### DIFF
--- a/CustomStartLayout.xml
+++ b/CustomStartLayout.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LayoutModificationTemplate
+    xmlns="http://schemas.microsoft.com/Start/2014/LayoutModification"
+    xmlns:defaultlayout="http://schemas.microsoft.com/Start/2014/FullDefaultLayout"
+    xmlns:start="http://schemas.microsoft.com/Start/2014/StartLayout"
+    xmlns:taskbar="http://schemas.microsoft.com/Start/2014/TaskbarLayout"
+    Version="1">
+  <LayoutOptions StartTileGroupCellWidth="6" StartTileGroupsColumnCount="1" />
+  <DefaultLayoutOverride>
+    <StartLayoutCollection>
+      <defaultlayout:StartLayout GroupCellWidth="6">
+      </defaultlayout:StartLayout>
+    </StartLayoutCollection>
+  </DefaultLayoutOverride>
+    <CustomTaskbarLayoutCollection PinListPlacement="Replace">
+      <defaultlayout:TaskbarLayout>
+        <taskbar:TaskbarPinList>
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%AppData%\Microsoft\Windows\Start Menu\Programs\System Tools\File Explorer.lnk"/>
+          <!--An Admin Command Prompt shortcut is created inside installer.vm-->
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%RAW_TOOLS_DIR%\Admin Command Prompt.lnk"/>
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%ProgramFiles%\Google\Chrome\Application\chrome.exe"/>
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%ProgramData%\chocolatey\bin\idafree.exe"/>
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%TOOL_LIST_DIR%\Networking\FakeNet-NG.lnk"/>
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%ProgramData%\chocolatey\bin\CFFExplorer.exe"/>
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%TOOL_LIST_DIR%\Utilities\procexp.lnk"/>
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%TOOL_LIST_DIR%\Utilities\procmon.lnk"/>
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%ProgramData%\chocolatey\bin\notepad++.exe"/>
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%TOOL_LIST_DIR%\Utilities\VisualStudio.lnk"/>
+        </taskbar:TaskbarPinList>
+      </defaultlayout:TaskbarLayout>
+    </CustomTaskbarLayoutCollection>
+</LayoutModificationTemplate>
--- a/install.ps1
+++ b/install.ps1
@@ -62,16 +62,41 @@
         https://github.com/mandiant/flare-vm
         https://github.com/mandiant/VM-Packages
 #>
+
 param (
   [string]$password = $null,
   [switch]$noPassword,
   [string]$customConfig = $null,
+  [string]$customLayout = $null,
   [switch]$noWait,
   [switch]$noGui,
   [switch]$noReboots,
   [switch]$noChecks
 )
 $ErrorActionPreference = "Stop"
+
+# Function used for getting configuration files (such as config.xml and CustomStartLayout.xml)
+function Get-ConfigFile {
+    param (
+        [string]$fileDestination,
+        [string]$fileSource
+    )
+    # Check if the source is an existing file path.
+    if (-not (Test-Path $fileSource)) {
+        # If the source doesn't exist, assume it's a URL and download the file.
+        Write-Host "[+] Downloading config file from '$fileSource'"
+        try {
+            (New-Object System.Net.WebClient).DownloadFile($fileSource, $fileDestination)
+        } catch {
+            Write-Host "`t[!] Failed to download '$fileSource'"
+            Write-Host "`t[!] $_"
+        }
+    } else {
+        # If the source exists as a file, copy it to the destination.
+        Write-Host "[+] Using existing file as configuration file."
+        Copy-Item -Path $fileSource -Destination $fileDestination -Force
+    }
+}
 
 # Set path to user's desktop
 $desktopPath = [Environment]::GetFolderPath("Desktop")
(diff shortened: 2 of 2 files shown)


install.ps1
```diff
@@ -840,6 +841,42 @@ foreach ($env in $configXml.config.envs.env) {
 }
 refreshenv
 
+Write-Host "[+] Checking for custom Start Layout file..."
+$customLayoutPath = Join-Path $desktopPath "CustomStartLayout.xml"
+if ([string]::IsNullOrEmpty($customLayout)) {
```
Ana06, line 846: The code we use to download `config.xml` and `CustomStartLayout.xml` is the same but using different url and filename. What about creating a helper function in this file (something like `Get-Config-File` with `$filePath` and `$url` as arguments) and use it for both functions? This way we don't have duplicated code and it should be easy to maintain the code.

install.ps1
```diff
@@ -840,6 +841,42 @@ foreach ($env in $configXml.config.envs.env) {
...
+        exit 1
+    }
+    if (-Not (Test-Path $customLayoutPath)) {
+        Write-Host "`t[!] Custom Start Layout file still missing: " $customLayoutPath -ForegroundColor Red
+        Write-Host "`t[!] Exiting..." -ForegroundColor Red
+        Start-Sleep 3
+        exit 1
+    }
+}
+
```
Ana06, line 879:
```suggestion
if (-Not (Test-Path $customLayoutPath)) {
    Write-Host "`t[!] Custom Start Layout file missing: " $customLayoutPath -ForegroundColor Red
    Write-Host "`t[-] No apps will be pinned to the taskbar. Do you want to continue? (Y/N): " -ForegroundColor Yellow -NoNewline
    $response = Read-Host
    if ($response -notin @("y","Y")) {
        exit 1
    }
}
```
I would remove this check. We need a `config.xml` file, but we don't necessarily need the `CustomStartLayout.xml` file. I would say if it does not exists, we just continue and don't apply any layout for the taskbar.

install.ps1
```diff
@@ -840,6 +841,42 @@ foreach ($env in $configXml.config.envs.env) {
 }
 refreshenv
 
+Write-Host "[+] Checking for custom Start Layout file..."
+$customLayoutPath = Join-Path $desktopPath "CustomStartLayout.xml"
```
Ana06, line 845: I suggest to write this file directly to `${Env:VM_COMMON_DIR}` or at least that we copy it there. For `config.xml`, we download it to the desktop first and then we save it to that directory (including additional packages possibly). I think the original reason to do it this way is that it is easier for the user to check if this file has been downloaded correctly (maybe @MalwareMechanic has more info on this). But as we display the path in the warning, I think this should be ok.

CustomStartLayout.xml
```diff
@@ -0,0 +1,30 @@
...
+    <StartLayoutCollection>
+      <defaultlayout:StartLayout GroupCellWidth="6">
+      </defaultlayout:StartLayout>
+    </StartLayoutCollection>
+  </DefaultLayoutOverride>
+    <CustomTaskbarLayoutCollection PinListPlacement="Replace">
+      <defaultlayout:TaskbarLayout>
+        <taskbar:TaskbarPinList>
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%AppData%\Microsoft\Windows\Start Menu\Programs\System Tools\File Explorer.lnk"/>
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%TOOL_LIST_DIR%\Utilities\Admin Command Prompt.lnk"/>
```
Ana06, line 19: [nitpick] We need this link because of the admin rights. I propose to use `%VM_COMMON_DIR%` or `RAW_TOOLS_DIR` for this file, as I don't think it is useful for the user in the `Utilities` folder. @mandiant/flare-vm opinions? Do you think this file is confusing in the Utilities folder or could it be useful?

CustomStartLayout.xml
```diff
@@ -0,0 +1,30 @@
...
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%ProgramFiles%\Google\Chrome\Application\chrome.exe"/>
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%ProgramData%\chocolatey\bin\notepad++.exe"/>
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%ProgramData%\chocolatey\bin\CFFExplorer.exe"/>
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%ProgramData%\chocolatey\bin\idafree.exe"/>
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%TOOL_LIST_DIR%\Utilities\procexp.lnk"/>
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%TOOL_LIST_DIR%\Utilities\procmon.lnk"/>
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%TOOL_LIST_DIR%\Networking\FakeNet-NG.lnk"/>
```
Ana06, line 26:
```suggestion
          <taskbar:DesktopApp DesktopApplicationLinkPath="%ProgramData%\chocolatey\bin\idafree.exe"/>
          <taskbar:DesktopApp DesktopApplicationLinkPath="%TOOL_LIST_DIR%\Networking\FakeNet-NG.lnk"/>
          <taskbar:DesktopApp DesktopApplicationLinkPath="%ProgramData%\chocolatey\bin\CFFExplorer.exe"/>
          <taskbar:DesktopApp DesktopApplicationLinkPath="%TOOL_LIST_DIR%\Utilities\procexp.lnk"/>
          <taskbar:DesktopApp DesktopApplicationLinkPath="%TOOL_LIST_DIR%\Utilities\procmon.lnk"/>
          <taskbar:DesktopApp DesktopApplicationLinkPath="%ProgramData%\chocolatey\bin\notepad++.exe"/>
```

Let's keep the order we are currently using for our builds (documented in #512)

Ana06 (changes requested):
This is a great approach to solve https://github.com/mandiant/flare-vm/issues/512, thanks for all the research and testing @emtuls

CustomStartLayout.xml
```diff
@@ -0,0 +1,30 @@
...
+        </taskbar:TaskbarPinList>
```
Ana06, line 27: As this approach will work once the path is available, we should also add Visual Studio (in case `visualstudio.vm` is installed, what we do in some of our VMs) to this list.

install.ps1
emtuls: Completed. 👍

install.ps1
emtuls: Completed. 👍

CustomStartLayout.xml
emtuls: Completed. 👍

install.ps1
emtuls: I added a copy command into my function so that if a custom path/file is provided, it will be copied to `${Env:VM_COMMON_DIR}` or if a URL is provided (or the default is used), it will be downloaded to there as well.

CustomStartLayout.xml
emtuls: I moved it to `RAW_TOOLS_DIR`, but I personally didn't mind it in Utilities. :)

CustomStartLayout.xml
emtuls: Completed! 👍

CustomStartLayout.xml
```diff
@@ -0,0 +1,31 @@
...
+    <StartLayoutCollection>
+      <defaultlayout:StartLayout GroupCellWidth="6">
+      </defaultlayout:StartLayout>
+    </StartLayoutCollection>
+  </DefaultLayoutOverride>
+    <CustomTaskbarLayoutCollection PinListPlacement="Replace">
+      <defaultlayout:TaskbarLayout>
+        <taskbar:TaskbarPinList>
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%AppData%\Microsoft\Windows\Start Menu\Programs\System Tools\File Explorer.lnk"/>
+          <taskbar:DesktopApp DesktopApplicationLinkPath="%RAW_TOOLS_DIR%\Admin Command Prompt.lnk"/>
```
Ana06, line 19:
```suggestion
          <taskbar:DesktopApp DesktopApplicationLinkPath="%RAW_TOOLS_DIR%\cmd.exe.lnk"/>
```

See https://github.com/mandiant/VM-Packages/pull/748#discussion_r1401745092

Ana06 (changes requested):
Thanks for the changes @emtuls, some minor things to address 😉 Looking forward to get this in and build a new VM! 🥳

CustomStartLayout.xml
emtuls: Leaving this as Admin Command Prompt to match Windows Dev VM implementation

Ana06: We discussed this outside GH, for documentation proposes: I agreed with @emtuls to leave it like it is to match Windows Dev VM implementation.

Ana06 (approved):
Great work @emtuls, looking forward to using this! 🥳 🍾 🎉

PR #534, opened 2023-11-22
config.xml: Add new packages to default config

Ana06:
Add recently created packages to the default configuration. @mandiant/flare-vm any objections? Any other package we are missing?

### DIFF
--- a/config.xml
+++ b/config.xml
@@ -48,6 +48,7 @@
         <package name="map.vm"/>
         <package name="nasm.vm"/>
         <package name="net-reactor-slayer.vm"/>
+        <package name="netcat.vm"/>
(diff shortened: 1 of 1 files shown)


config.xml
```diff
@@ -59,6 +60,7 @@
         <package name="processdump.vm"/>
         <package name="reg_export.vm"/>
         <package name="regshot.vm"/>
+        <package name="rpcview.vm"/>
```
Ana06, line 63: I am going to remove `rpcview.vm` because it seems that is breaking in the daily run, likely because a dependency requirement being incompatible with another package dependency requirement. We can add it until we have fixed that.

mr-tz (approved):

PR #536, opened 2023-11-22
Fix: Removing comment from CustomStartLayout.xml

emtuls:
Comments seem to make the taskbar pinning completely break. The powershell cmdlet to parse and verify file integrity does not complain, so this may be a bug/oversight on the Microsoft side.


Ana06 (approved):
Good catch @emtuls!

[nitpick] Limit the commit message title (first line) to 50 characters so that it displays nicely in the GH interface and other tools that rely on this convention:
Screenshot 2023-11-23 at 09 01 30

PR #551, opened 2023-12-09
[Enhancement] Better install.ps1 network checks

HuskyHacks:
Hello again!

I've noticed a few PMAT students (particularly our analyst friends in countries other than the US) report install.ps1 failing network checks. The script currently doesn't give a lot of info about how the network checks failed. This PR seeks to provide more info so an analyst can troubleshoot more effectively.

All checks are still encapsulated so the `-noChecks` bypass will skip them:

image

Running without a custom config and all checks succeed:

image

Example of the ping check failing:

image

Erroring out on the GET request to GitHub:

image

When running `install.ps1` while pulling a remote custom config file, the check will run for raw.githubusercontent.com:

image

Network checks are now encapsulated in a simple function, so if we want to add more of them (choco site, etc) it's as simple as:

```powershell
    Test-WebConnection "google.com"
    Test-WebConnection "github.com"
    # Test-WebConnection "[any other URL we need]"
```

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -75,6 +75,39 @@ param (
 )
 $ErrorActionPreference = "Stop"
 
+# Function to test the network stack. Ping/GET requests to the resource to ensure that network stack looks good for installation
+function Test-WebConnection {
+    param (
+        [string]$url
+    )
+
+    Write-Host "[+] Checking for Internet connectivity ($url)..."
+
+    if (-not (Test-Connection $url -Quiet)) {
+        Write-Host "`t[!] It looks like you cannot ping $url. Check your network settings." -ForegroundColor Red
+        Start-Sleep 3
+        exit 1
+    }
+
+    $response = $null
+    try {
+        $response = Invoke-WebRequest -Uri "https://$url" -UseBasicParsing
+    }
+    catch {
+        Write-Host "`t[!] Error accessing $url. Exception: $($_.Exception.Message)`n`t[!] Check your network settings." -ForegroundColor Red
+        Start-Sleep 3
+        exit 1
+    }
+
+    if ($response -and $response.StatusCode -ne 200) {
+        Write-Host "`t[!] Unable to access $url. Status code: $($response.StatusCode)`n`t[!] Check your network settings." -ForegroundColor Red
+        Start-Sleep 3
+        exit 1
+    }
+
+    Write-Host "`t[+] Internet connectivity check for $url passed" -ForegroundColor Green
+}
+
 # Function used for getting configuration files (such as config.xml and CustomStartLayout.xml)
 function Get-ConfigFile {
     param (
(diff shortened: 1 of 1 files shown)


install.ps1
```diff
@@ -75,6 +75,41 @@ param (
...
+
+    Write-Host "[+] Checking for Internet connectivity ($url)..."
+
+    if (-not (Test-Connection $url -Quiet)) {
+        Write-Host "`t[!] It looks like you cannot access $url. Check your network settings." -ForegroundColor Red
+        Start-Sleep 3
+        exit 1
+    }
+
+    Write-Host "`t[+] Ping check for $url passed" -ForegroundColor Green
```
Ana06, line 92: I would remove this line because the internet check is getting too noisy. I think one message saying the check succeeded at the end of the function is enough.

install.ps1
```diff
@@ -222,20 +257,18 @@ if (-not $noChecks.IsPresent) {
...
-    else
-    {
-        Write-Host "`t[+] Internet connectivity detected" -ForegroundColor Green
+
+    # Internet connectivity checks
+    Test-WebConnection "google.com"
+    Test-WebConnection "github.com"
+    
+    # Checking for raw.githubusercontent.com connectivity if the URL is in the custom config
+    if ($customConfig -and $customConfig.Contains("raw.githubusercontent.com")) {
```
Ana06, line 266: We download the background image from `raw.githubusercontent.com`. So I think we should remove this `if` and test it always.

Ana06 (changes requested):
Thanks for your contribution @HuskyHacks!

Could you squash your commits in order to keep a clean commit history? Otherwise we can merge the PR with `squash and merge`.

HuskyHacks:
Apologies for the wacky rebase, my formatter went crazy on the rest of the file but I've restored it to its previous state! We should be good now.

install.ps1
```diff
@@ -75,6 +75,39 @@ param (
...
+        exit 1
+    }
+
+    if ($response -and $response.StatusCode -ne 200) {
+        Write-Host "`t[!] Unable to access $url. Status code: $($response.StatusCode)`n`t[!] Check your network settings." -ForegroundColor Red
+        Start-Sleep 3
+        exit 1
+    }
+
+    Write-Host "`t[+] Web check for $url passed" -ForegroundColor Green
```
Ana06, line 108: [nitpick] Should we remove the `Web` here?

Ana06 (commented):
Thanks for the changes @HuskyHacks!

install.ps1
HuskyHacks: "Internet connectivity check passed" work here?

Ana06 (approved):
Thanks @HuskyHacks!

PR #556, opened 2023-12-22
Add the needed 2024 Q1 packages

d35ha:


Ana06 (approved):
Thanks for adding these useful tools to our default configuration @d35ha! 🎉

PR #560, opened 2024-01-15
[config] Add blobrunner to default config

Ana06:
Add 32-bit and 64-bit versions of blobrunner to the default FLARE-VM configuration. These packages are added in the following PRs (which need to be merged before this one):
- https://github.com/mandiant/VM-Packages/pull/836
- https://github.com/mandiant/VM-Packages/pull/835

@mandiant/flare-vm any other package that should be added to the default config?


emtuls (approved):
Awesome! Great tools to use for sure. :)

Issue #559, opened 2024-01-12
Link CyberChef instead of GoogleChrome to taskbar


Ana06:
### Details

Link CyberChef instead of Google Chrome to taskbar or open CyberChef as default main page in Google Chrome. This requires modifying flare-vm and/or VM-Packages.

Issues that may be related: https://github.com/mandiant/VM-Packages/issues/823, https://github.com/mandiant/VM-Packages/issues/822, https://github.com/mandiant/VM-Packages/issues/725, https://github.com/mandiant/VM-Packages/issues/296

emtuls:
1. Chrome seems to change where they put their user preferences for homepages often enough that I don't think programmatically adjusting the preferences file to set the default homepage is going to be a good option for long term (but we can consider it if need be).

2. Attempting to set the homepage to a local file (the `cyberchef .html file`) via `chrome://policy` did not work. It seems to prohibit the use of local files.

3. I think merging the fixes for these 2 issues will give me the ground I need to make a good alternative (using a command prompt shortcut): 
- https://github.com/mandiant/VM-Packages/issues/842
- https://github.com/mandiant/VM-Packages/issues/843

Ana06:
We need to merge https://github.com/mandiant/flare-vm/issues/559 to close this (I modified the description so that it happens automatically)

emtuls:
I think you might mean https://github.com/mandiant/flare-vm/pull/562? 🙂

PR #562, opened 2024-01-17
Pin cyberchef shortcut instead of chrome

emtuls:
This attempts to fix https://github.com/mandiant/flare-vm/issues/559

To complete it, we will also need to merge: https://github.com/mandiant/VM-Packages/pull/844

Closes https://github.com/mandiant/flare-vm/issues/559


Ana06 (approved):
Thanks @emtuls!

emtuls:
The CLA has been running for almost 24 hours...Not sure what we should do. Seems bugged. 😕

Ana06:
I don't know how we can re-run it. We could re-push to force it to re-run. But I think we can just merge the PR without the CLA check, we know it is fine as there are only commits from you.

emtuls:
I attempted to rerun it once after it did that for ~2 hours and it still ended up doing the same thing...Thanks for merging!

PR #568, opened 2024-02-19
[config] Add packages to default configuration

Ana06:
Anything against adding these packages to the default configuration? Any other package we should add?

### DIFF
--- a/config.xml
+++ b/config.xml
@@ -31,6 +31,7 @@
         <package name="dotnet3.5"/> <!-- To run old .NET binaries -->
         <package name="explorersuite.vm"/>
         <package name="fakenet-ng.vm"/>
+        <package name="file.vm"/>
         <package name="floss.vm"/>
         <package name="ghidra.vm"/>
         <package name="goresym.vm"/>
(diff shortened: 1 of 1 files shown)


binjo (approved):
LGTM

emtuls:
Looks good! I have a couple of additions, but they can always be added at a later time. 😊

- Depending on the route we want to take for 7z, we can swap out the old 7z with `7z-nsis` once that gets merged in?
- `TotalRegistry` seems like a good option to add for a default. 
- `Microsoft-Windows-Terminal` might also be good.

emtuls (approved):

Ana06:
@emtuls 
> - Depending on the route we want to take for 7z, we can swap out the old 7z with `7z-nsis` once that gets merged in?
>- `TotalRegistry` seems like a good option to add for a default. 
>- `Microsoft-Windows-Terminal` might also be good. 

Thanks for sharing your ideas! `microsoft-windows-terminal.vm` is already in the default config. It was added in https://github.com/mandiant/flare-vm/commit/3c5b131cf6e15f93669ee7ce533fb21cdb1740b0. @emtuls I agree with the other two suggestions.

@mandiant/flare-vm opinions on adding TotalRegistry to the default config? Note a similar tool (RegCool) has also been added.

Issue #582, opened 2024-04-05
Taskbar Shortcuts Doubling Up


emtuls:
### What's the problem?

Currently, `CFF Explorer` and `notepad++` will have a shortcut in the taskbar, but when clicked, it will show a second icon next to it for the running program, likely due to these using the chocolately shim shortcuts.

Now that we have an actual shortcut for `notepad++`, we can use that one directly, and also fix `CFF Explorer` to use the proper shortcut, and this mitigates the issue.

### Steps to Reproduce

Run flarevm installer and test `CFF Explorer` and `notepad++` shortcuts.

### Environment

environment agnostic

PR #583, opened 2024-04-05
Improve taskbar shortcut links

emtuls:
This fixes https://github.com/mandiant/flare-vm/issues/582

Currently, `CFF Explorer` and `notepad++` will have a shortcut in the taskbar, but when clicked, it will show a second icon next to it for the running program, likely due to these using the chocolately shim shortcuts.

Now that we have an actual shortcut for `notepad++`, we can use that one directly, and also fix `CFF Explorer` to use the proper shortcut, and this mitigates this issue.


Ana06 (approved):
Thanks @emtuls!

PR #584, opened 2024-04-12
Add more useful tools to Flare default

emtuls:


config.xml
```diff
@@ -39,6 +41,7 @@
         <package name="hashmyfiles.vm"/>
         <package name="hollowshunter.vm"/>
         <package name="hxd.vm"/>
+        <package name="ida.diaphora.vm"/>
```
Ana06, line 44: I like this tool and would like it in the VM. But it just copies the file into the VM instead of installing it as plugin. I thought this may be confusing and that's the reason it hasn't been added to the default config before. There is an issue to explore installing it as a plugin: https://github.com/mandiant/VM-Packages/issues/994 Should we wait for that or would you prefer to add the current version now?

Ana06 (commented):
What about adding `pdbresym.vm` from https://github.com/mandiant/VM-Packages/pull/993 as well? (we can do it in a separate PR too)

@mandiant/flare-vm any other package we should add?

config.xml
emtuls: I think adding it now is fine, since it isn't effecting anything and gives the user the ability to play with it in its current state while also allowing it to be updated in the future if the package changes, without us needing to modify this config again. :)

Ana06: we would need to change the config as the name of the package would change, but is not a big thing. :smile:

Ana06 (approved):

config.xml
```diff
@@ -59,15 +63,18 @@
...
         <package name="pebear.vm"/>
         <package name="peid.vm"/>
         <package name="pesieve.vm"/>
         <package name="pestudio.vm"/>
         <package name="pkg-unpacker.vm"/>
         <package name="pma-labs.vm"/>
+        <package name="procdot.vm"/>
         <package name="processdump.vm"/>
         <package name="reg_export.vm"/>
+        <package name="regcool.vm"/>
```
Ana06, line 77: We should remove regcool as it is broken at the moment: https://github.com/mandiant/VM-Packages/issues/1004

Ana06 (changes requested):

Ana06 (approved):
I have removed regcool.vm and kept the rest.

PR #587, opened 2024-04-22
[README] Add password expiration

Ana06:
A password expiration blocks the FLARE-VM installation as it asks to change the password after a restart. I had this issue today. Document how to set `Password never expires` in the user properties in `lusrmgr.msc` to avoid the issue.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -229,10 +229,8 @@ if (-not $noChecks.IsPresent) {
 
     # Check for spaces in the username, exit if identified
     Write-Host "[+] Checking for spaces in the username..."
-    $currentUsername = [System.Security.Principal.WindowsIdentity]::GetCurrent().Name
-    $extractedUsername = $currentUsername -replace '^.*\\'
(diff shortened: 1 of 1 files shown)


emtuls:
Is this something that we could automate and do early during the installation process? 

It seems to just require running `WMIC USERACCOUNT WHERE Name='<username here>' SET PasswordExpires=FALSE`, but I could be wrong.

Ana06:
@emtuls 
> It seems to just require running WMIC USERACCOUNT WHERE Name='<username here>' SET PasswordExpires=FALSE, but I could be wrong.

have you tried this? It doesn't work for me:
image

emtuls:
Whoops. That is how you would do it in `Command Prompt`. 
It would be a little different if doing it from Powershell: `Set-LocalUser -Name "<username here>" -PasswordNeverExpires $true `

**Command Prompt**
image

**Powershell**
image

emtuls (approved):
Looks good! Thanks for making this happen!

PR #589, opened 2024-04-23
Fix fakenet shortcut pinning to taskbar

### LINKED ISSUE #588: Fakenet no longer pinned to taskbar
### What's the problem?

We recently changed the tool name for FakeNet from `FakeNet-NG` to `fakenet`, but we did not update the `CustomStartupLayout.xml` to also reflect this change, causing it to not be pinned during FlareVM installation,

### Steps to Reproduce

Install FlareVM with FakeNet as a selected package.

### Environment

Environment Agnostic.

### Additional Information

_No response_

emtuls:
This fixes https://github.com/mandiant/flare-vm/issues/588


Ana06 (approved):
Thanks @emtuls!

PR #590, opened 2024-04-23
[CustomStartLayout.xml] Fix VisualStudio

Ana06:
We have changed visualstudio.vm of category breaking the shortcut in the taskbar.


emtuls (approved):
Good to go 🙂

PR #604, opened 2024-08-12
Added improvements to reduce errors and such

CamoCatX:
I added 4 different changes to this script.
1. On powershell 5.1 (comes with default installation of windows), there is a significant time-burner for Invoke-Webrequest  and other commands. It comes from writing that light blue progress bar at the top of the window. More info here: https://github.com/PowerShell/PowerShell/issues/2138 
2. My next improvement is handling if a user accidently clicks on PowerShell x86. The script then corrects this human error.
3. Disable Keep alive for invoke web request is just because we should not assume that user will have further interaction beyond what is required on the script, so we should remove for potential less latency.
4. I added Microsoft.com instead of Google because of how likely of each to be blocked. This is also for personal privacy reasons, this features adds nothing extra but on a windows machine is usually preferred. I chose not to do a more privacy respecting website because that might also be blocked in a weird corporate environment.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -76,7 +76,16 @@ param (
   [switch]$noReboots,
   [switch]$noChecks
 )
-$ErrorActionPreference = "Stop"
+$ErrorActionPreference = 'Stop'
+$ProgressPreference = 'SilentlyContinue'
+ 
+# https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.localaccounts/set-localuser
+if ( -not ([Environment]::Is64BitProcess)) {
+    Write-Host "`t[!] It seems like you are using 32 bit powershell. Exiting because some commands do not work correctly in this mode." -ForegroundColor Red
+    Write-Host "`t[-] Hint: Don't run powershell x86" -ForegroundColor Yellow
+    Start-Sleep 3
+    exit 1
+}
 
 # Function to test the network stack. Ping/GET requests to the resource to ensure that network stack looks good for installation
 function Test-WebConnection {
@@ -94,7 +103,7 @@ function Test-WebConnection {
 
     $response = $null
     try {
-        $response = Invoke-WebRequest -Uri "https://$url" -UseBasicParsing
+        $response = Invoke-WebRequest -Uri "https://$url" -UseBasicParsing -DisableKeepAlive
     }
     catch {
         Write-Host "`t[!] Error accessing $url. Exception: $($_.Exception.Message)`n`t[!] Check your network settings." -ForegroundColor Red
@@ -257,9 +266,9 @@ if (-not $noChecks.IsPresent) {
     }
 
     # Internet connectivity checks
-    Test-WebConnection "google.com"
-    Test-WebConnection "github.com"
-    Test-WebConnection "raw.githubusercontent.com"
+    Test-WebConnection 'microsoft.com'
+    Test-WebConnection 'github.com'
+    Test-WebConnection 'raw.githubusercontent.com'
 
     Write-Host "`t[+] Network connectivity looks good" -ForegroundColor Green
 
--- a/install.ps1
+++ b/install.ps1
@@ -266,9 +266,9 @@ if (-not $noChecks.IsPresent) {
     }
 
     # Internet connectivity checks
-    Test-WebConnection 'microsoft.com'
-    Test-WebConnection 'github.com'
     Test-WebConnection 'google.com'
+    Test-WebConnection 'github.com'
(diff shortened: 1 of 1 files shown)


install.ps1
```diff
@@ -76,7 +76,16 @@ param (
   [switch]$noReboots,
   [switch]$noChecks
 )
-$ErrorActionPreference = "Stop"
+$ErrorActionPreference = 'Stop'
+$ProgressPreference = 'SilentlyContinue'
+ 
+# https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.localaccounts/set-localuser
+if ( -not ([Environment]::Is64BitProcess)) {
+    Write-Host "`t[!] It seems like you are using 32 bit powershell. Exiting because some commands do not work correctly in this mode." -ForegroundColor Red
```
Ana06, line 84: As powershell x86 is not working at the moment, I think adding this code is a good idea. But on the long-term, is there a reason why we would want to support powershell x86 (we would need to find alternative implementations for the used function(s))? :thinking: @mandiant/vms opinions?

install.ps1
```diff
@@ -94,7 +103,7 @@ function Test-WebConnection {
 
     $response = $null
     try {
-        $response = Invoke-WebRequest -Uri "https://$url" -UseBasicParsing
+        $response = Invoke-WebRequest -Uri "https://$url" -UseBasicParsing -DisableKeepAlive
```
Ana06, line 106: Nice improvement, this function is only used to test the internet connection so we do not need a persistent connection to the server as there are not subsequent requests.

install.ps1
```diff
@@ -257,9 +266,9 @@ if (-not $noChecks.IsPresent) {
     }
 
     # Internet connectivity checks
-    Test-WebConnection "google.com"
-    Test-WebConnection "github.com"
-    Test-WebConnection "raw.githubusercontent.com"
+    Test-WebConnection 'microsoft.com'
```
Ana06, line 269: > I added Microsoft.com instead of Google because of how likely of each to be blocked. This is also for personal privacy reasons, this features adds nothing extra but on a windows machine is usually preferred. I chose not to do a more privacy respecting website because that might also be blocked in a weird corporate environment.

Testing `google.com` is one of most (if not the most) used internet connectivity check. I see no reason why a corporate environment would block it. This looks to me like a personal preference and not an improvement of the code, so I would prefer to keep it as it was. Other opinions @mandiant/flare-vm?

@CamoCatX can you please remove this change so that we can merge this PR?

Ana06 (changes requested):
Thanks for the PR @CamoCatX, nice improvements! :sparkles: 

Can you remove the `google.com` replacement so that we can merge this PR? :smile: 

[nitpick] I would appreciate a more descriptive commit message, suggestion:

```
[installer] Improve install.ps1 code

- Remove progress bar for better performance
- Check for PowerShell x86 as it is unsupported
- Disable `KeepAlive` in connectivity check as no subsequent request are
  expected
```

CamoCatX:
@Ana06 I have changed this back to google, even though I personally know many people who would prefer Microsoft if in a windows environment. But it should be able to be merged now.

CamoCatX:
If I have free time in the future I can look into making this script compatible with pwsh x86.

install.ps1
Ana06, line 269: This change is now removing `raw.githubusercontent.com`

Ana06 (changes requested):
Thanks for the modification, but I think you accidentally changed the wrong domain. Please restore the domains in the internet connectivity check so that we can merge the PR.

I would prefer if you could squash your commits (to avoid fix commits in order to keep a clean commit history), but we can also merge the PR with `squash an merge` if you prefer.

install.ps1
CamoCatX: Sorry for the minor typo, this is fixed.

Ana06 (approved):
Thanks for the improvements @CamoCatX! 

> If I have free time in the future I can look into making this script compatible with pwsh x86. The problem I found when using x86 was with the Set-LocalUser cmdlet. Microsoft documentation lists this: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.localaccounts/set-localuser To resolve this issue we would have to find an alternative to the Microsoft.PowerShell.LocalAccounts module

I have created https://github.com/mandiant/flare-vm/issues/615 for this. It would be great if could work on it! :smile: If/When you have time to work on it, just write a comment in the issue to avoid someone else works on it at the same time. :wink:

PR #605, opened 2024-08-21
Add virtualbox folder with scripts

Ana06:
Add virtualbox folder with scripts related to enhance building, exporting, and using FLARE-VM in VirtualBox and a README. Adding the following two scripts in this PR, but planning to continue contributing to this folder:
- `vbox-clean-snapshots.py` that cleans a VirtualBox VM up by deleting a snapshot and its children recursively skipping snapshots with a substring in the name. Its is not possible to select and delete several snapshots in the VirtualBox UI, making cleaning up your VM manually after having creating a lot snapshots time consuming and tedious.
- `vbox-export-snapshots.py` exports one or more snapshots in the same VirtualBox VM as `.ova`, changing the network adapter to Host-Only. The script also generates a file with the SHA256 hash of the exported `.ova`. This script is useful to export several versions of FLARE-VM after its installation consistently and with the internet disabled by default (desired for malware analysis).


virtualbox/vbox-clean-snapshots.py
```diff
@@ -0,0 +1,94 @@
...
+    if argv is None:
+        argv = sys.argv[1:]
+
+    epilog = textwrap.dedent(
+        """
+        Example usage:
+          # Delete all snapshots that do not include 'clean' or 'done' in the name (case insensitive) in the 'FLARE-VM.20240604' VM
+          vbox-clean-snapshots.py FLARE-VM.20240604
+
+          # Delete the 'CLEAN with IDA 8.4' snapshot an its children recursively skipping the ones that include 'clean' or 'done' in the name (case insensitive) in the 'FLARE-VM.20240604' VM
```
emtuls, line 65: Minor typo - "'CLEAN with IDA 8.4' snapshot an its children", should be *and*

virtualbox/README.md
```diff
@@ -0,0 +1,58 @@
+# VirtualBox scripts
+
+**This folder contains several scripts related to enhance building, exporting, and using FLARE-VM in VirtualBox.**
+
+## Clean up snapshots
+
+Its is not possible to select and delete several snapshots in VirtualBox, making cleaning up your virtual machine (VM) manually after having creating a lot snapshots time consuming and tedious (possible errors when deleting several snapshots simultaneously).
```
emtuls: Minor typo - "Its is not possible to select and delete", first word should just be "It"

virtualbox/vbox-clean-snapshots.py
emtuls: Also, this example slightly confuses me. Does

emtuls (changes requested):
Minor typos to fix, but outside of that, awesome stuff! Really excited to have these scripts. :)

virtualbox/vbox-clean-snapshots.py
Ana06: in general it would delete the provided snapshot and its children for those whose name don't include the protected words. In this case, you are right that because `clean` is on the name, only the children could be deleted. What about replacing this case by the following two to clarify it?
```
# Delete the 'CLEAN with IDA 8.4' children snapshots recursively skipping the ones that include 'clean' or 'done' in the name (case insensitive) in the 'FLARE-VM.20240604' VM
vbox-clean-snapshots.py FLARE-VM.20240604 --root_snapshot 'CLEAN with IDA 8.4'

# Delete the 'Snapshot 3' snapshot an its children recursively skipping the ones that include 'clean' or 'done' in the name (case insensitive) in the 'FLARE-VM.20240604' VM
vbox-clean-snapshots.py FLARE-VM.20240604 --root_snapshot 'Snapshot 3'
```

emtuls: I think we should keep both, but maybe reword the first one a little bit. How does this sound?
```
# Delete the 'CLEAN with IDA 8.4' children snapshots recursively skipping the ones that include 'clean' or 'done' in the name (case insensitive; **NOTE:** the root snapshot is skipped in this case) in the 'FLARE-VM.20240604' VM
vbox-clean-snapshots.py FLARE-VM.20240604 --root_snapshot 'CLEAN with IDA 8.4'
```

Ana06: Thanks for the improvement suggestions @emtuls!

emtuls (approved):

PR #608, opened 2024-08-27
Add internet detector tool images

emtuls:
This adds the images required by https://github.com/mandiant/VM-Packages/pull/1121


install.ps1
```diff
@@ -924,6 +924,14 @@ $backgroundImage = "${Env:VM_COMMON_DIR}\background.png"
 # Use background image for lock screen as well
 $lockScreenImage = "${Env:VM_COMMON_DIR}\lockscreen.png"
 Copy-Item $backgroundImage $lockScreenImage
+# Download FLARE VM (internet enabled) background image
+$backgroundInternetImage = "${Env:VM_COMMON_DIR}\background-internet.png"
+(New-Object net.webclient).DownloadFile('https://raw.githubusercontent.com/mandiant/flare-vm/main/Images/flarevm-background-internet.png', $backgroundInternetImage)
+# Download internet detection tool indicator icons
```
Ana06, line 930: I think the icons should be part of the tool or the tool package, as they shouldn't be configurable as the background image

install.ps1
```diff
@@ -924,6 +924,14 @@ $backgroundImage = "${Env:VM_COMMON_DIR}\background.png"
 # Use background image for lock screen as well
 $lockScreenImage = "${Env:VM_COMMON_DIR}\lockscreen.png"
 Copy-Item $backgroundImage $lockScreenImage
+# Download FLARE VM (internet enabled) background image
+$backgroundInternetImage = "${Env:VM_COMMON_DIR}\background-internet.png"
```
Ana06, line 928: I wonder if it is possible to create a `VM` subfolder in `Images` and download the whole folder content to `${Env:VM_COMMON_DIR}` instead of downloading all of them individually like you are doing now, it allows us more flexibility if we need to change the images names or add more images. we would still need to duplicate the lockscreen image.

emtuls:
Unfortunately, I don't think we can go the route of simply downloading a directory off of github with the images we want all inside. :(
It either needs to be packaged into a single file (like a zip), or we have to iterate through each file we want like we had before.

Someone does maintain a tool that can sort of make it happen, but idk if we want to rely on that or if it works easily (I think it might require a github token, so probably not something we want to consider): https://github.com/download-directory/download-directory.github.io

Ana06:
We  can download https://github.com/mandiant/flare-vm/archive/refs/heads/main.zip and extract the Images/VM folder in a small function. I think it is nicer that downloading individual images.

Ana06:
I think we should split this PR in one to add the images and one to add the downloading of the images to `install.ps1`

Ana06 (approved):
Thanks @emtuls! :+1:

PR #609, opened 2024-08-28
[virtualbox] Add vbox-adapter-check.py

Ana06:
Add script to print the status of all internet adapters of all VMs in VirtualBox. The script notifies using PyGObject if any dynamic analysis VM (with `.dynamic` in the name) has an adapter whose type is not allowed (internet access is often undesirable for dynamic malware analysis). Optionally change the type of the adapters with non-allowed type to Host-Only.

The script has been tested in Debian 12 with GNOME 44.9. Please let me know if you test it in a different environment.

I have included in this PR a commit with small improvements in the `vbox-clean-snapshots.py` documentation.
#### Notification
Screenshot from 2024-08-28 11-03-22

#### Notification list 
Screenshot from 2024-08-28 11-05-20

#### Notification list with the [_Expandable Notifications_ extension](https://extensions.gnome.org/extension/4463/expandable-notifications/)
Screenshot from 2024-08-28 11-03-48


virtualbox/vbox-adapter-check.py
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/python3
+
+import sys
+import textwrap
+import argparse
+import virtualbox
+from virtualbox.library import NetworkAttachmentType as NetType
+from gi.repository import Notify
```
williballenthin, line 8: What are the dependencies for this script? Where are they specified?

Typically this would go in a `requirements.txt` file, though with a bunch of disparate scripts, keeping them in the file header can also work. Recently, I've been migrating to using `uv run` to invoke single-file scripts with dependencies in the header: https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies

virtualbox/vbox-adapter-check.py
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/python3
```
williballenthin, line 1: missing copyright header

virtualbox/vbox-adapter-check.py
Ana06: the only dependency that shouldn't be already installed in Debian with GNOME is `virtualbox`.

williballenthin: also, it may be worth catching `ImportError`s and explaining to the user when things don't work:

  - for `virtualbox`: what they need to do to install the bindings
  - for `gi`: that the script must be run on a system with Gnome

Ana06: >it may be worth catching `ImportError`s and explaining to the user when things don't work:

I like this idea. I also like the idea of the `requirements.txt` file, but it is a bit tricky with the current structure as all the scripts are in the same folder an they could have different dependencies. :thinking: 

>   - for `virtualbox`: what they need to do to install the bindings

The `virtualbox` library is required by all the scripts in the `virtualbox` folder. so if we add this, we should add it to all the scripts for consistency

>   - for `gi`: that the script must be run on a system with Gnome

I think `gi` may work on other types of systems as well, but I am not sure in which. I have tried this notification in openSUSE with KDE and it seems to work.

virtualbox/vbox-adapter-check.py
Ana06: sorry, there is one single file with such header, the linter that @mr-tz contributed: [scripts/test/lint.py](https://github.com/mandiant/VM-Packages/blob/921ab89f37a2513fa77571f695415b2160970b03/scripts/test/lint.py#L8)

williballenthin: adding copyright headers to all files and a LICENSE.txt would bring the project closer in line to Google recommendations, but should be its own effort, not this PR.

virtualbox/vbox-adapter-check.py
```diff
@@ -0,0 +1,94 @@
...
+
+          # Print status of enabled internet adapters and disabled the enabled adapters with internet access in VMs with {DYNAMIC_VM_NAME} in the name
+          vbox-adapter-check.vm --skip_disabled
+
+          # # Print status of enabled internet adapters without modifying any of them
+          vbox-adapter-check.vm --skip_disabled --do_not_modify
+        """
+    )
+    parser = argparse.ArgumentParser(
+            description=f"Print the status of all internet adapters of all VMs in VirtualBox. Notify if any VM with {DYNAMIC_VM_NAME} in the name has an adapter whose type is not allowed (internet access is undesirable for dynamic malware analysis)i. Optionally change the type of the adapters with non-allowed type to Host-Only.",
```
emtuls, line 78: [nitpick]: `description` argument doesn't seem to line up with the other 2 arguments. Should we align them all?

emtuls (approved):
Generally overall good stuff! I like it a lot and the new notification is much nicer than our previous prototype!

I agree with Willi that it would be nice to have a print in a catch that explains what a user may need to do in order to fix an issue with import, but not 100% necessary. Just a nice to have. :)

Thanks for the hard work @Ana06!

virtualbox/vbox-adapter-check.py
Ana06: I have been looking into adding this and I think we should leave it out, the error looks like:
```
Traceback (most recent call last):
  File "/home/anamg/github/flare-vm/virtualbox/./vbox-adapter-check.py", line 7, in <module>
    import virtualbox
ModuleNotFoundError: No module named 'virtualbox'
```
You can install this library the same way as any other Python library (`python3 -m pip install virtualbox`), so I think a custom message does not help more than the generic error for which you can find help online. @williballenthin are you ok with merging it as it is? We can discuss how to improve if needed afterwards.

PR #612, opened 2024-09-02
[config] Add internet_detector.vm

Ana06:
Add `internet_detector.vm` introduced in https://github.com/mandiant/VM-Packages/pull/1121 to the default configuration. Note that the tool is not running by default at the moment because of the noise in fakenet: https://github.com/mandiant/flare-fakenet-ng/issues/190 but we are considering [running the tool by default with a scheduled task](https://github.com/mandiant/VM-Packages/blob/24c8bcadf3a89d99401936ca6c49ffbfd6f199df/packages/internet_detector.vm/tools/chocolateyinstall.ps1#L28).

image

### DIFF
--- a/config.xml
+++ b/config.xml
@@ -56,6 +56,7 @@
         <package name="ilspy.vm"/>
         <package name="innoextract.vm"/>
         <package name="innounp.vm"/>
+        <package name="internet_detector.vm"/>
         <package name="isd.vm"/>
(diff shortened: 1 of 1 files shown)


emtuls (approved):

emtuls:
Awesome! Glad we got that all together! Thank you @Ana06 !

PR #614, opened 2024-09-05
[vbox-clean-snapshots] Lowercase comparison

Ana06:
Lowercase both the snapshot name and the protected snapshot substrings to use lowercase comparison.


binjo (approved):
LGTM

Issue #603, opened 2024-08-07
Error with packages installer.vm


monfora:
### What's the problem?

At the end of installation packages to the system there is an error with package.

### Steps to Reproduce

clean iso of windows 11 from microsoft -> new vm on virtualbox -> default selected packages on install stage

### Environment

- VirtualBox
- Windows 11 - 10.0.22631
- 5.1.22621.3880
- 2.3.0
- Boxstarter|3.0.3

### Additional Information

[terminal_logs.txt](https://github.com/user-attachments/files/16757470/terminal_logs.txt)
[chocolatey.log](https://github.com/user-attachments/files/16757473/chocolatey.log)
[chocolatey.1.log](https://github.com/user-attachments/files/16757474/chocolatey.1.log)
[chocolatey.2.log](https://github.com/user-attachments/files/16757475/chocolatey.2.log)
[log.txt](https://github.com/user-attachments/files/16757476/log.txt)

Ana06:
With the provided information it is not clear to me which package fails and what the error is. Please provide the console output as required in the bug issue template so that we are able to help you.

emtuls:
Thanks for reporting this @monfora!

@Ana06 I think I see the issue here. It looks like the `CustomStartLayout.xml` is not being applied and this seems to be an issue that others on newer Windows 11 builds are also experiencing: https://www.reddit.com/r/SCCM/comments/1cus7g3/importstartlayout_for_taskbar_stopped_working/

It seems that Microsoft may have either changed the format spec (though I can't find an updated one on the internet), or they are disallowing this ability via PowerShell (some people say their old configs still work when applied via GPO).

Fortunately, it looks like there is a simple fix, which is to store the layout as `LayoutModification.xml` and place it in `C:\Users\default\AppData\Local\Microsoft\Windows\Shell\`.

I just tested the fix on a Windows 10 pre-built FlareVM and it worked fine. I'll need to get a proper new Windows 11 build set up to test it there. I also need to test it from a fresh ISO through the full build to see if there may be any other tweaking that is needed (my Win 11 test should help with that as well).

To test, once the file is named correctly and placed in that location (and assuming you've removed any other pinned icons), you can run `Stop-Process -Name explorer -Force` and it should work. But this will need to be adjusted in our repo and `installer.vm` if it is the route we decide to take.

monfora:
fixed by adding terminal log

monfora:
> Thanks for reporting this @monfora!
> 
[...]

PR #616, opened 2024-09-13
First step to fix Windows 11 install failure

emtuls:
This is a 2 fold PR.

1. First step for addressing the installation failure in Windows 11 as noted here: https://github.com/mandiant/flare-vm/issues/603. This changes the name of our custom start layout to the name that Windows uses when actually implementing the custom layout, which will then be used in (2nd PR here) to store it in the correct location.
2. This also fixes Procmon and Process Explorer duplication in the taskbar as noted in https://github.com/mandiant/VM-Packages/issues/1100

**NOTE**: This WILL break the install until the 2nd PR is merged due to the name change for the custom layout.


emtuls:
Closed in favor of doing everything in a single PR: https://github.com/mandiant/flare-vm/pull/617

PR #617, opened 2024-09-13
Fix Custom Taskbar from breaking on Win 11

emtuls:
This is a 2 fold PR.

1. This is the first step for addressing the installation failure in Windows 11 as noted here: https://github.com/mandiant/flare-vm/issues/603. This changes the name of our custom start layout to the name that Windows uses when actually implementing the custom layout, which will then be used in (2nd PR [here](https://github.com/mandiant/VM-Packages/pull/1137), from VM-Packages) to store it in the correct location.

2. This also fixes Procmon and Process Explorer duplication in the taskbar as noted in https://github.com/mandiant/VM-Packages/issues/1100

**NOTE:** This WILL break the install until the 2nd PR is merged due to the name change for the custom layout.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -120,7 +120,7 @@ function Test-WebConnection {
     Write-Host "`t[+] Internet connectivity check for $url passed" -ForegroundColor Green
 }
 
-# Function used for getting configuration files (such as config.xml and CustomStartLayout.xml)
+# Function used for getting configuration files (such as config.xml and LayoutModification.xml)
 function Get-ConfigFile {
     param (
         [string]$fileDestination,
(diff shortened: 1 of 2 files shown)


binjo (approved):

install.ps1
```diff
@@ -910,9 +910,9 @@ $configXml.save((Join-Path ${Env:VM_COMMON_DIR} "packages.xml"))
 
 # Custom Start Layout setup
 Write-Host "[+] Checking for custom Start Layout file..."
-$layoutPath = Join-Path ${Env:VM_COMMON_DIR} "CustomStartLayout.xml"
+$layoutPath = Join-Path "C:\Users\Default\AppData\Local\Microsoft\Windows\Shell" "LayoutModification.xml"
```
Ana06, line 913: @emtuls 

> This changes the name of our custom start layout to the name that Windows uses when actually implementing the custom layout, which will then be used in (2nd PR https://github.com/mandiant/VM-Packages/pull/1137, from VM-Packages) to store it in the correct location.

I do not see that the new filename is used in the second PR. Am I understanding something wrong?

Ana06: good, I just wanted to confirm it before merging this PR :wink:

Ana06 (approved):

PR #630, opened 2024-11-25
Force DNS requests to always come from requesting process

emtuls:
This works in tandem with https://github.com/mandiant/VM-Packages/pull/1178 so that internet_detector will be blocked properly within FakeNet using Process Block List.


Ana06 (approved):
It seems there are no concerns :smile:

PR #638, opened 2025-01-13
Improve license + copyright & add CONTRIBUTING file

Ana06:
- Replace LICENSE text file with the official Apache one: https://www.apache.org/licenses/LICENSE-2.0.txt Note the license is not being changed.
- Add copyright headers to source code files to follow Google conventions.
- Add CONTRIBUTING file with information about CLA and Google's code of conduct to follow Google conventions/policies and make contributing easier. Move the information of the _Contributing_ section of the README to this document.

PR #641, opened 2025-01-17
Fix for Issue #640 - config.xml misformatted

### LINKED ISSUE #640: Broken default config.xml
### What's the problem?

PR #638 added a comment to the default config.xml which seems to be interfering with the install script.
With it, the install will fail completely, early in the setup process.

### Steps to Reproduce

1. Follow install instructions
2. Observe that install.ps1 fails with this error:

Image

### Environment

VMware Workstation
10.0.19045
5.1.19041.5369
2.4.1
Boxstarter|3.0.3

### Additional Information

_No response_

thomasklemenc:
- Move the license after the opening XML tag. Per spec, first characters of an XML file must be `<?xml`
> ... encoding must begin with an XML encoding declaration, in which the first characters must be '<?xml' ...
- Remove the BOM as it is not needed for UTF-8 encoded XML files

Fixes #640

CLA has been signed.

### DIFF
--- a/config.xml
+++ b/config.xml
@@ -1,3 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
 <!--
  Copyright 2017 Google LLC
 
@@ -14,7 +15,6 @@
  limitations under the License.
 -->
 
-﻿<?xml version="1.0" encoding="utf-8"?>
 <config>
     <envs>
         <env name="VM_COMMON_DIR" value="%ProgramData%\_VM"/>


Saad-Alaref (approved):
I have done the same thing before reading this pull request, and it does indeed work.

Ana06 (approved):
It seems I broke this file in https://github.com/mandiant/flare-vm/pull/638, breaking FLARE-VM :see_no_evil: I used [addlicense](https://github.com/google/addlicense) to add the headers and it seems a `<feff>` (Byte Order Mark (BOM)) in this file caused that it added the header in the wrong way, breaking the file. I am not sure if this a bug in [addlicense](https://github.com/google/addlicense) as the BOM likely shouldn't be there, as I also think it is not needed. In any case, @tGecko thanks for reporting and fixing the bug! :bouquet: @Saad-Alaref thank for confirming the fix work and the additional details :bouquet:

JackStuart:
Hi @Ana06, It was a bug in Addlicence - I have created a PR to fix it. Maybe you might have more traction getting someone in Google to take a look at it because it hasn't been updated in 2.5 years

https://github.com/google/addlicense/pull/166

PR #645, opened 2025-01-23
[vbox-clean-snapshots.py] Fix several bugs & improve code

Ana06:
The following PR fixes the following **bugs**:
- **A snapshot outside root may be deleted if it has the same name as one inside root**.Use the snapshot ID instead of the name to fix the issue.
-  **The root snapshot argument should be optional and case sensitive**.  Make clear that this is the case and use all snapshots if no root snapshot is provided instead of raising an exception.
- **Duplicated current snapshot**. The current code may include the current snapshot twice, which causes that
a confusing output is rendered and that the snapshot is being removed twice (failing the second time).
- **Root snapshot is not deleted if a child is protected, but it is possible to delete a root snapshot with a single protected child.** It is not possible to delete the root snapshot if it has more than one protected child. But this issue happens at every level, not only at the root snapshot. The inconsistent behavior is confusing and complicates the code unnecessarily. Instead, try to always delete non protected snapshots (including the root snapshot) and improve the rendered error output.
- **Protected root snapshot is not supported**. The root snapshot can be protected, so that its children get deleted but not the root snapshot. 

In addition, the PR includes other improvements, such as better exception handling and using more specific VBoxManage commands. **See the detailed commit messages for more details on every of the changes**.

### DIFF
--- a/virtualbox/vbox-clean-snapshots.py
+++ b/virtualbox/vbox-clean-snapshots.py
@@ -58,9 +58,7 @@ def get_snapshot_children(vm_name, root_snapshot_name, protected_snapshots):
         # find the root SnapshotName by matching the name
         root_snapshotid = None
         for snapshotid, snapshot_name in snapshots:
-            if snapshot_name.lower() == root_snapshot_name.lower() and (
-                not any(p.lower() in snapshot_name.lower() for p in protected_snapshots)
-            ):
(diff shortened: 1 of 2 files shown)


williballenthin (approved):
beautiful sequence of commits and messages. thank you!

stevemk14ebr:
+1 the commit messages are :chefs_kiss: I can do a quick once over tomorrow

PR #647, opened 2025-01-28
Improve vboxcommon

Ana06:
Improve `vboxcommon.py`, including:
- Enhance, make consistent and remove duplication in `ensure_vm_running` and `ensure_vm_shutdown` by introducing `wait_until_vm_state`. Remove the waiting time of 2 minutes after calling `ensure_vm_running` from `ensure_vm_shutdown` when the state is `saved, as `ensure_vm_running` already waits for the state to change to `running`.
- Enhance `ensure_hostonlyif_exists` by introducing `get_hostonlyif_name` to avoid duplication.
- Parse the VBoxManage output with regular expressions for consistency and to improve code readability. 
- Remove noisy output and make the output consistent to make easier to call vboxcommon functions from other scripts displaying a nice output.
- Remove unneeded use of exceptions re-raising.

Also rename  variables in `vbox-adapter-check.py` and `vboxcommon.py` for consistency with VBoxManage and the other vbox scripts.

See the detailed commit messages for more details.

### DIFF
--- a/virtualbox/vboxcommon.py
+++ b/virtualbox/vboxcommon.py
@@ -101,72 +101,50 @@ def get_vm_state(vm_uuid):
     raise Exception(f"Could not start VM '{vm_uuid}'")
 
 
-def ensure_vm_running(vm_uuid):
-    """Checks if the VM is running and starts it if it's not.
-    Waits up to 1 minute for the VM to transition to the 'running' state.
-    """
-    try:
-        vm_state = get_vm_state(vm_uuid)
-        if vm_state != "running":
-            print(
-                f"VM {vm_uuid} is not running (state: {vm_state}). Starting VM..."
-            )
-            run_vboxmanage(["startvm", vm_uuid, "--type", "gui"])
+def wait_until_vm_state(vm_uuid, target_state):
+    """Wait for VM state to change.
 
-            # Wait for VM to start (up to 1 minute)
-            timeout = 60  # seconds
-            check_interval = 5  # seconds
-            start_time = time.time()
-            while time.time() - start_time < timeout:
-                vm_state = get_vm_state(vm_uuid)
-                if vm_state == "running":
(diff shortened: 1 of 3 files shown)


virtualbox/vboxcommon.py
```diff
@@ -101,72 +101,50 @@ def get_vm_state(vm_uuid):
...
 def ensure_vm_shutdown(vm_uuid):
-    """Checks if the VM is running and shuts it down if it is."""
-    try:
-        vm_state = get_vm_state(vm_uuid)
-        if vm_state == "saved":
-            print(
-                f"VM {vm_uuid} is in a saved state. Powering on for a while then shutting down..."
-            )
-            ensure_vm_running(vm_uuid)
-            time.sleep(120)  # 2 minutes to boot up
```
stevemk14ebr, line 148: Still need this to ensure the guest is fully booted to the desktop before initiating a powerdown.

stevemk14ebr (changes requested):

virtualbox/vboxcommon.py
Ana06: We are just starting the VM here because we can't shut it down in a saved state. Once the state is running, the shutdown works without any extra waiting. Or am I missing something?

I have tested this code with several saved states (saving it before it is booted to the desktop for example) and this code works for all of them:
```
./vbox-export-snapshots.py 
Starting operations on FLARE-VM
VM {2bc66f50-9ecb-4b10-a4dd-0cc329bc383d} state: saved. Starting VM...
VM {2bc66f50-9ecb-4b10-a4dd-0cc329bc383d} state: running. Shutting down VM...
```

stevemk14ebr: If there's a wait in the vbox-export-snapshots I think that would be fine. I recall from testing I did have to wait a short while for it to work well.

PR #649, opened 2025-01-30
Improve vbox-export-snapshots.py

Ana06:
Improve `vbox-export-snapshots.py`, including:
- **Continue exporting snapshots if one fails to export**
- **Receive the configuration information as arguments in `vbox-export-snapshots.py` instead of hardcode it in global variables**. This allows more flexibility when using the script without having to modify it. The command gets a bit long though. Let me know if you have any improvement idea. :wink: 
- **Do a power cycle to ensure everything is good and wait for 4 minutes to give the internet detector time to detect the network change**. A similar code was before in `ensure_vm_running`, but the waiting time only happened if the VM was in state `saved` when starting the shutdown. In addition, it unnecessarily delayed other functions that call this function. This is the change discussed with @stevemk14ebr in https://github.com/mandiant/flare-vm/pull/647#discussion_r1933386823.
- **Get the VM UUID and create the export directory only once** (instead of for every snapshot)
- **Move the calls to `ensure_vm_shutdown` inside the functions that require the VM to be shutdown** instead of having it as a pre-step for better code readability and preventing issues.
- **Improve output and error handling**.
- **Improve code readability and documentation**.

# Example

### After
With the changes in https://github.com/mandiant/flare-vm/pull/647 and in this PR.  Note we export an extra snapshot after one fails to export.

```
$ ./vbox-export-snapshots.py "FLARE-VM.testing" --snapshot "FLARE-VM,.dynamic,Windows 10 VM with FLARE-VM default configuration" --snapshot "FLARE-VM.full,.full.dynamic,Windows 10 VM with FLARE-VM default configuration + 'visualstudio.vm' + 'pdbs.pdbresym.vm'" --snapshot "FLARE-VM.EDU,.EDU,Windows 10 VM with FLARE-VM default configuration + FLARE-EDU materials"

Exporting snapshots from "FLARE-VM.testing" {2bc66f50-9ecb-4b10-a4dd-0cc329bc383d}
Export directory: "/home/anamg/EXPORTED VMS"

VM {2bc66f50-9ecb-4b10-a4dd-0cc329bc383d} ✨ restored snapshot "FLARE-VM"
VM {2bc66f50-9ecb-4b10-a4dd-0cc329bc383d} state: saved. Starting VM...
VM {2bc66f50-9ecb-4b10-a4dd-0cc329bc383d} state: running. Shutting down VM...
VM {2bc66f50-9ecb-4b10-a4dd-0cc329bc383d} ⚙️  network set to single hostonly adapter
[...]

PR #652, opened 2025-02-03
[config] Remove microsoft-office.vm & add ida.plugin.hrtng.vm

Ana06:
- Remove microsoft-office.vm as it only provides 5 days trial and makes it confusing when installing another Office version. It can still be installed using a custom config file or selecting it via the GUI.
- Add the recently created package ida.plugin.hrtng.vm for the hrtng IDA plugin.

### DIFF
--- a/config.xml
+++ b/config.xml
@@ -86,7 +86,6 @@
         <package name="magika.vm"/>
         <package name="malware-jail.vm"/>
         <package name="map.vm"/>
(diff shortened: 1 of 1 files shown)


Ana06:
we can merge it even the linter is failing. https://github.com/mandiant/flare-vm/pull/649 fixes the linter.

Issue #615, opened 2024-09-05
 Support Powershell x86


Ana06:
### Details

We use at lease one function that is not available in Powershell x86: [Set-LocalUser](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.localaccounts/set-localuser). In https://github.com/mandiant/flare-vm/pull/604, we added an error message an exit the installation if Powershell x86 is used.

It would be better to support Powershell x86 by finding alternative implementations for the incompatible function(s).

PR #658, opened 2025-02-09
Support Powershell x86 Issue #615 - Alternate Implementation of Set-LocalUser

PrajeetGuha:
I have added the alternate implementation of the `Set-LocalUser` and hence the code is working in both powershell x86 and x64 without any issue. I have used WMIC functionality to do the same activity of making the password of the user to never expire.  Since the x86 related issue was limited to that only function, I have also removed the check condition of the powershell.

Closes https://github.com/mandiant/flare-vm/issues/615

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -94,14 +94,6 @@ param (
 )
 $ErrorActionPreference = 'Stop'
 $ProgressPreference = 'SilentlyContinue'
- 
-# https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.localaccounts/set-localuser
-if ( -not ([Environment]::Is64BitProcess)) {
-    Write-Host "`t[!] It seems like you are using 32 bit powershell. Exiting because some commands do not work correctly in this mode." -ForegroundColor Red
-    Write-Host "`t[-] Hint: Don't run powershell x86" -ForegroundColor Yellow
-    Start-Sleep 3
-    exit 1
-}
 
 # Function to test the network stack. Ping/GET requests to the resource to ensure that network stack looks good for installation
 function Test-WebConnection {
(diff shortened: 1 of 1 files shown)


PrajeetGuha:
Thanks for your reply @Ana06 . I am actually facing a problem with the CLA. I have actually authored the commit with my local PC which was configured with a random non-existent mail id as authoring information. Since I am new to open source contributions, I did not know it will cause this issue. I also could not find any way to change or resolve the issue.
Could you suggest anyway to resolve it?😅
I have signed the CLA with this GH account as evident from #1273

Ana06:
@PrajeetGuha you can change the commit author of the last commit with:
```
git commit --amend --author="Author Name <email@gmail.com>" --no-edit
```

You can change the global git configuration (so that it is applied for all future commits) with:
```
git config --global user.name "Author Name"
git config --global user.email "email@gmail.com"
```

Let me know if this work. :blush:

PrajeetGuha:
Thanks @Ana06 for the help. I did the changes. Looks like it passed the CLA check now.
image

Ana06 (approved):
Thanks @PrajeetGuha!

install.ps1
```diff
@@ -340,7 +332,7 @@ if (-not $noChecks.IsPresent) {
     }
 
     Write-Host "[+] Setting password to never expire to avoid that a password expiration blocks the installation..."
-    Set-LocalUser -Name  "${Env:UserName}" -PasswordNeverExpires $true
+    wmic useraccount here "name='${Env:UserName}'" set PasswordExpires=False
```
sara-rn, line 335: there is a typo here, it should be `where`

PR #660, opened 2025-02-12
Automated Creation and Export of Customized FLARE-VM Builds

Ana06:
# `vbox-build-flare-vm.py`: Automated Creation and Export of Customized FLARE-VM Builds

This PR enhances `vbox-build-flare-vm.py` (previously named `vbox-build-vm.py`) to fully automate the creation and export of customized FLARE-VM builds. The script orchestrates the entire process, including:
-   Restoring a pre-existing `BUILD-READY` snapshot of a clean Windows installation (with UAC disabled).
- Copying required installation files (such as the IDA Pro installer,  the FLARE-VM configuration file, and legal notices) into the guest VM.
- Installing FLARE-VM and creating a `base` snapshot.
- Generating subsequent snapshots and exporting OVA images based on a YAML configuration file.

## Configuration File

The YAML configuration file specifies the VM name, exported VM name, and details for each snapshot. Snapshot configurations support custom commands, legal notices, and file/folder exclusions for automated cleanup. See `virtualbox/configs/win10_flare-vm.yaml` for an example to export three OVAs:
- `.dynamic`: Installs IDA Pro and Microsoft Office 2024 in addition to the FLARE-VM default configuration
- `.full.dynamic`:  Downloads Windows Symbols and installs IDA Pro, Microsoft Office 2024 and Visual Studio in addition to the FLARE-VM default configuration. This OVA exhibits a significant size increase compared to `.dynamic`.
- `.EDU`: Unzips a ZIP archive containing educational materials (labs and demos) and installs Microsoft Office 2016.

## `--do-not-install-flare-vm` Flag 
When the `--do-not-install-flare-vm` flag is provided, `vbox-build-flare-vm.py` skips the initial FLARE-VM installation and assumes the existence of a pre-existing base snapshot. This enables modification of the base snapshot (e.g., to reinstall failed packages) created by a prior run and resumption of the build process from the adjusted base snapshot.

## Handling the VirtualBox `VERR_NO_LOW_MEMORY` Bug

VirtualBox encounters a `VERR_NO_LOW_MEMORY` error due to a bug (https://www.virtualbox.org/ticket/22185) that causes incorrect calculation of available memory. To prevent this issue from disrupting the script, when the bug is detected within `run_vboxmanage`, fix instructions are printed, and the command is retried after a brief delay.

## Robuster `ensure_vm_running` Implementation

Previously, `ensure_vm_running` only verified if the VM was in the `running` state, which is insufficient as the guest OS might still be booting. Introduce the `get_num_logged_in_users` function and use it in `ensure_vm_running` to wait until at least one user is logged in. However, even this is not always enough in slower environments,, as the VirtualBox guest additions can take a while to load after user login. Therefore, in addition a retry mechanism with a 2-minute delay has been implemented for `VBoxManage guestcontrol` commands that fail.

## Handling the `aborted-saved` State in `ensure_vm_shutdown`

`ensure_vm_shutdown` now avoids attempting to shut down the VM if its state is `aborted-saved` (e.g. after a VM crash), as this operation fails because the VM is not running.

## Preventing OVA Export Conflicts

Export failures due to existing OVA files are prevented by renaming the old OVA adding a timestamp.

## `vbox-export-snapshots.py` Simplification for Single Snapshot Export

Initially, `vbox-export-snapshots.py` was intended to export FLARE-VM snapshots built by `vbox-build-flare-vm.py`.However, this proved problematic because `vbox-build-flare-vm.py` requires executing `VM-Set-Legal-Notice`, which prevents the VM from starting without manual interaction (to accept the legal notice). Starting the VM is necessary after setting the network interface to `hostonly` for the internet detector to update the VM status. Therefore, `hostonly` setup and snapshot export were moved into `vbox-build-flare-vm.py`. With `vbox-export-snapshots.py` now decoupled from `vbox-build-flare-vm.py`, `vbox-export-snapshots.py` has been simplified to export single snapshots, eliminating  the complexity of managing multiple snapshots and their associated configuration files. Shared code was moved to `vboxcommon.py` to avoid duplication.
### Completely Unattended Process TODOs

**For a fully automated, unattended process that requires no manual intervention, some issues encountered during testing still need to be addressed** (these are not part of this PR).

#### Ensuring `failed_packages.txt` Existence

The script relies on the `failed_packages.txt` file to determine FLARE-VM installation completion. Currently, this file is only created if any package installation fails. `malware-jail.vm` currently fails, albeit the tool is installed (see https://github.com/mandiant/VM-Packages/issues/1130). This results in at least one failing package and the existence of `failed_packages.txt`. To ensure continued reliance on this file after the `malware-jail.vm` package is fixed, it must always be created. This is implemented in https://github.com/mandiant/VM-Packages/pull/1274.

#### Package Installation Failures

Occasionally, packages fail to install, and a simple retry would resolve the issue. Currently, this is handled manually using the `--do-not-install-flare-vm` flag explained above. To eliminate manual intervention, package installation retries on failure during FLARE-VM installation are necessary: https://github.com/mandiant/VM-Packages/issues/1219

#### VM Crashes

I have experienced VM crashes occasionally during FLARE-VM installation (likely due to VM restarts), particularly with newer Debian kernels. The current implementation of `vbox-build-flare-vm.py` takes snapshots every ~20 minutes, which can be restored in case of a crash. However, this restoration is currently manual. Until the VM is manually restored to a pre-crash snapshot, the script continues to take snapshots of the crashed VM. The script needs to be extended to:
-   Detect VM crashes.
-   Copy the VirtualBox log to `LOGS_DIR` for crash debugging.
-   Restart a previous snapshot before the crash to resume installation.

I'll create an issue to track this.

#### Other Proposed Enhancements

The following proposed changes would further enhance script execution:
-   Create subfolders within `LOGS_DIR` with installation date and time in their names instead of deleting previous logs, facilitating debugging of past runs.
-   Reduce noisy output from certain commands within the build VM script, such as `VM-Clean-Up`, to improve readability of `vbox-build-flare-vm.py` output.

I'll create issues to track these enhancement proposals.


stevemk14ebr (approved):

stevemk14ebr:
What is the target path of the broken link. Does it break after the first subpath with a space? Do you have a path escape issue?

Ana06:
> What is the target path of the broken link. Does it break after the first subpath with a space? Do you have a path escape issue?

There is not target path in the broken link. The package code that creates the shortcut is the same if I run the PS command from the host with the script or from the guest: https://github.com/mandiant/VM-Packages/blob/e9d0d073a1759890db3221c509940917fda42ca9/packages/visualstudio.vm/tools/chocolateyinstall.ps1#L17 So it should be created in the exactly same way.

Ana06:
I have also experience the issue with Word tools, where the shortcuts are either missing or broken.
https://github.com/mandiant/VM-Packages/blob/e9d0d073a1759890db3221c509940917fda42ca9/packages/microsoft-office.vm/tools/chocolateyinstall.ps1#L24

PR #670, opened 2025-03-13
[config] Add nmap.vm

Ana06:
Add nmap.vm to the default configuration which installs nmap, ncat, ndiff, and nping.

We removed netcat.vm in favor of nmap.vm (as it includes ncat) in https://github.com/mandiant/flare-vm/pull/633. But it seems we forgot to add nmap.vm to the default configuation. I missed ncat today. :sob:


sara-rn (approved):
looks good to me

Issue #578, opened 2024-03-20
Display packages by category in installer


jstrosch:
### Details

Hey all! Have you considered creating a series of custom configurations that could align to "profiles"? For example, a profile that installs just the tools needed for PE file analysis or malicious documents? This could help reduce install time by helping folks avoid installing packages they don't need. I realize there is a way to customize the install, but these profiles can be passed via command line (XML) to help with anyone that wants/needs more automation (or doesn't want to have to manually tweak the install each time). I've created a few as examples at: https://github.com/jstrosch/FLARE-VM-configs/tree/main. Curious to get everyone's thoughts?

Ana06:
@emtuls recently shared a similar idea with me. I also discussed it with @MalwareMechanic some time ago. But it seems we never created an issue and we didn't have time to work on it. Thanks for documenting the idea @jstrosch!

heap-s:
@Ana06 Is this something that is being considered still?

emtuls:
@heap-s Definitely still being considered though it is unfortunately not on our high priority list at the moment.

We do have a plan to work on improving the GUI interface, primarily adding categories and drop down lists to the package list, and if time permits, we would try to squeeze that in as well. :)

Ana06:
I think organizing tool by categories in the installer UI and allowing to select/unselect all of them would be a nice improvement that addresses this issue and https://github.com/mandiant/flare-vm/issues/432. My understanding is that @emtuls is planing to implement this after implementing https://github.com/mandiant/VM-Packages/issues/1107 as we need a good way to access the categories in the installer. @emtuls I'll assign you the issues to ensure others are aware you are working on it.

necrose99:
winget ui / [unigetui ](https://github.com/marticliment/UniGetUI)  one can make a json apps install  to get custom pkg adds choco / winget pypi scoop etc..

Ana06:
@necrose99 can you elaborate on how that would exactly work? Can we customize the UI with a FLARE-VM theme and the packages to use our feed?

Ana06:
@sara-rn is planing to modify the installer to use https://www.myget.org/F/vm-packages/Packages after she adds the category as [tag in the nuspec](https://learn.microsoft.com/en-us/nuget/reference/nuspec#tags) to display the packages by category and to allow to select a whole category at once. @sara-rn is researching the best way to modify the installer to implement this.

Ana06:
[...]

Issue #546, opened 2023-11-27
Remove installer.vm and deboat.vm from package customization GUI


emtuls:
### Details

`installer.vm` and `debloat.vm` are both automatically installed during the `flare-vm` installation inside of `install.ps1`, so they are not necessary to be displayed inside of the GUI interface for package options to select from.

Ana06:
I agree, we should remove then from the installer GUI. Would you like to address this @emtuls?

Ana06:
This issue can be addressed together with https://github.com/mandiant/flare-vm/issues/578

Issue #432, opened 2022-12-24
Organize Available Packages in Installer GUI


MalwareMechanic:
As packages continue to increase, it'd be great if we had a way to better display / organize available packages.

Initial ideas:

- By category
- By distribution flavor (e.g., FLARE VM, Commando VM, Threat Pursuit VM)

emtuls:
I'm planning to take a stab at improving the FlareVM GUI packages interface and sorting things into categories.

My initial attempt has led me to realizing that we may need a slight design change somewhere, so that I'm able to get categories (and ideally their associated packages) over to the flare-vm installer easily. 
* I wish we could pass more metadata through our CI via `choco push` to our `myget` so that when we pull the info via `choco search`, we'd be able to get the associated category as well, but that seems to be unlikely, since that's not what it's designed for.

My current though is to modify `categories.txt` so that we include the associated package names with each category using an `xml` structure instead. 
* This would require us to keep it in sync with any category changes (either at a package level or just category name level), but I think this tradeoff is worth the investment. 
* This also gives us a quick high level overview of what packages fall into each category at a glance, which might be useful data to be aware of.
* This would also of course require some modifications to various templates and scripts that rely on categories.txt the way it is now, which should be fairly trivial imo.

PR #671, opened 2025-03-14
Add PowerShell linter & minor enhacement in the installer

Ana06:
- Add PowerShell linter similar to the [linter in VM-Packages](https://github.com/mandiant/VM-Packages/blob/main/scripts/test/lint.ps1). Fix offenses detected by the linter manually, including replacing the WMI cmdlet by the CIM cmdlet. Closes https://github.com/mandiant/flare-vm/issues/507.
- Check that Windows is >= 10 instead of only checking `6.1.7601`.
- Add missing user name to info message and write the environment variable name always in the same way. This issue was introduced when I simplified the code to get the user name in https://github.com/mandiant/flare-vm/commit/cc103449a543cc7452bbb2aa8612549eed240595.

PR #672, opened 2025-03-14
Update installer to organize packages by categories

sara-rn:
Update installer so that after the initial form to set up the environmental variables a new window displays all the packages grouped by category. 
- The packages from `config.xml` need to be selected by default.
- The option to select all the packages from a specific category, for example if the category is a checkbox, doesn't work well as the checkboxes need to be generated dynamically. Currently each category is displayed in a label.

Closes https://github.com/mandiant/flare-vm/issues/578
Closes https://github.com/mandiant/flare-vm/issues/546
Closes https://github.com/mandiant/flare-vm/issues/432

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -481,7 +481,8 @@ if (-not $noGui.IsPresent) {
     }
 
     function Get-Packages-Categories {
-     $vmPackagesUrl = "https://www.myget.org/F/vm-packages/api/v2/Packages?$filter=IsLatestVersion%20eq%20true"
+     # MyGet API
+     $vmPackagesUrl = "https://www.myget.org/F/vm-packages/api/v2/Packages"
      $vmPackagesFile = "${Env:VM_COMMON_DIR}\vm-packages.xml"
      $packagesByCategory=@{}
      do {
@@ -504,18 +505,26 @@ if (-not $noGui.IsPresent) {
 
         # Extract package information from the XML
         $vm_packages.SelectNodes("//atom:entry", $ns) | ForEach-Object {
-            $isLatestVersion = $_.SelectSingleNode("m:properties/d:IsLatestVersion", $ns).InnerText
-            if ($isLatestVersion -eq "true") {
-                $packageName = $_.SelectSingleNode("m:properties/d:Id", $ns).InnerText
+             $isLatestVersion = $_.SelectSingleNode("m:properties/d:IsLatestVersion", $ns).InnerText
+             $category = $_.SelectSingleNode("m:properties/d:Tags", $ns).InnerText
+
+             #we select only packages that have the latest version and contain a category
+             if ($isLatestVersion -eq "true" -and $category) {
+                  $packageName = $_.SelectSingleNode("m:properties/d:Id", $ns).InnerText
                   $description = $_.SelectSingleNode("m:properties/d:Description", $ns).InnerText
-                  $category = $_.SelectSingleNode("m:properties/d:Tags", $ns).InnerText
-                  if ($category -ne ""){
-                    if (-not ($packagesByCategory.ContainsKey($category))) {
-                     $packagesByCategory[$category] = ""
-                    }
-                    $packagesByCategory[$category] += $packageName +  ":" + $description + "`n"
+
+                  if (-not ($packagesByCategory.ContainsKey($category))) {
+                     # Initialize as an empty array
+                     $packagesByCategory[$category] = @()
                   }
-             }
+                  # category should not be empty, this condition should be removed after all the old packages in nuget are removed
+                  if ($category -ne ""){
+                    $packagesByCategory[$category] += [PSCustomObject]@{
+                      PackageName = $packageName
+                      PackageDescription = $description
+                      }
+                  }
+              }
         }
         # Check if there is a next link in the XML and set the API URL to that link if it exists
         $nextLink = $vm_packages.SelectSingleNode("//atom:link[@rel='next']/@href", $ns)
@@ -769,17 +778,7 @@ if (-not $noGui.IsPresent) {
         param (
          [string]$category
         )
-        $packages = @()
-        $lines = $packagesByCategory[$category] -split "`n"
-        foreach ($line in $lines) {
-            if ($line.Trim()){
-               $package, $description = $line -split '\|'
-			   $packages += [PSCustomObject]@{
-                        PackageName   = $package
-                        PackageDescription = $description
-                    }
-            }
-        }
+        $packages = $packagesByCategory[$category]
         return $packages
     }
 
@@ -822,8 +821,8 @@ if (-not $noGui.IsPresent) {
     $allPackagesButton.Font            = New-Object System.Drawing.Font('Microsoft Sans Serif',10)
     $allPackagesButton.Add_Click({Select-AllPackages})
 
-    $clearPackagesButton			   = New-Object system.Windows.Forms.Button
-    $clearPackagesButton.text            = "Select All"
+    $clearPackagesButton	         = New-Object system.Windows.Forms.Button
+    $clearPackagesButton.text            = "Deselect All"
     $clearPackagesButton.AutoSize        = $true
     $clearPackagesButton.location        = New-Object System.Drawing.Point(210,750)
     $clearPackagesButton.Font            = New-Object System.Drawing.Font('Microsoft Sans Serif',10)
@@ -877,9 +876,6 @@ if (-not $noGui.IsPresent) {
 			$checkBox.Font = New-Object System.Drawing.Font('Microsoft Sans Serif',9)
 			$checkBox.AutoSize = $true
 			$checkBox.Location = New-Object System.Drawing.Point(40, $verticalPosition2)
-			if ($package.PackageName -in $packagesToInstall){
-			   $checkBox.Checked = $true
-			}
 			$checkBox.Name = "checkBox$numCheckBoxPackages"
 			$checkboxesPackages.Add($checkBox)
 			$Panel_Categories.Controls.Add($checkBox)
@@ -889,6 +885,10 @@ if (-not $noGui.IsPresent) {
 		$verticalPosition += 20 * ($NumPackages ) + 30  # Increment to space checkboxes vertically
 		$numCategories ++
 	}
+
+    #select packages that are in the config.xml
+    Set-InitialPackages
+
 
 	$FormCategories.controls.AddRange(@($categories_label,$Panel_Categories,$ContinueButton,$resetButton,$allPackagesButton,$cancelButton,$clearPackagesButton))
 	$FormCategories.Add_Shown({$FormCategories.Activate()})
--- a/install.ps1
+++ b/install.ps1
@@ -461,6 +461,26 @@ if (-not $noGui.IsPresent) {
     ################################################################################
     Add-Type -AssemblyName System.Windows.Forms
 
+    function Get-Packages-Categories {
+      $categoriesFile = "${Env:VM_COMMON_DIR}\categories.txt"
+      $categoriesUrl = "https://raw.githubusercontent.com/mandiant/VM-Packages/refs/heads/main/categories.txt"
+      (New-Object net.webclient).DownloadFile($categoriesUrl, $backgroundImage)
+
+       Get-Content -Path $categoriesFile | ForEach-Object {
+         $currentCategory = $_
+         $packages = Find-Package -Source https://www.myget.org/F/vm-packages/ -FilterOnTag $currentCategory
+         $packagesByCategory[$currentCategory] = ""
+         Foreach ($package in $packages) {
+                $packageName = $package.Name
+                $description = $package.Summary
+                $packagesByCategory[$currentCategory] += $packageName +  "|" + $description + "`n"
+                #Write-Host "[+] Added the " $packageName " to the category " $currentCategory " with description " $description
+         }
+       }
+       return $packagesByCategory
+
+    }
+
     function Get-Folder($textBox, $envVar) {
         $folderBrowserDialog = New-Object System.Windows.Forms.FolderBrowserDialog
         $folderBrowserDialog.RootFolder = 'MyComputer'
@@ -493,116 +513,35 @@ if (-not $noGui.IsPresent) {
             }
         }
     }
-
-    function Set-InitialPackages {
-        $selectedPackagesBox.Items.Clear()
-        foreach($package in $packagesToInstall)
-        {
-            $selectedPackagesBox.Items.Add($package) | Out-Null
-        }
-        $numSelectedLabel.text = "Total: $($selectedPackagesBox.Items.count)"
-
-        $unselectedPackagesBox.Items.Clear()
-        foreach($package in $allPackages)
-        {
-            $unselectedPackagesBox.Items.Add($package) | Out-Null
-        }
-        $numUnselectedLabel.text = "Total: $($unselectedPackagesBox.Items.count)"
-    }
-
-    function Add-SelectedPackages {
-        $unselectedPackages = $unselectedPackagesBox.SelectedItems
-        foreach($package in $unselectedPackages)
-        {
-            $selectedPackagesBox.BeginUpdate()
-            $selectedPackagesBox.Items.Add($package) | Out-Null
-            $selectedPackagesBox.EndUpdate()
-        }
-        $numSelectedLabel.text = "Total: $($selectedPackagesBox.Items.count)"
-
-        $unselectedPackagesBox.BeginUpdate()
-        while ($unselectedPackagesBox.SelectedItems.count -gt 0) {
-            $unselectedPackagesBox.Items.RemoveAt($unselectedPackagesBox.SelectedIndex)
-        }
-        $unselectedPackagesBox.EndUpdate()
-        $numUnselectedLabel.text = "Total: $($unselectedPackagesBox.Items.count)"
-    }
-
-    function Add-AllPackages {
-        foreach($package in $unselectedPackagesBox.Items)
-        {
-            $selectedPackagesBox.BeginUpdate()
-            $selectedPackagesBox.Items.Add($package) | Out-Null
-            $selectedPackagesBox.EndUpdate()
-        }
-        $numSelectedLabel.text = "Total: $($selectedPackagesBox.Items.count)"
-
-        $unselectedPackagesBox.BeginUpdate()
-        $unselectedPackagesBox.Items.Clear()
-        $unselectedPackagesBox.EndUpdate()
-        $numUnselectedLabel.text = "Total: $($unselectedPackagesBox.Items.count)"
-    }
-
-    function Remove-SelectedPackages {
-        $selectedPackages = $selectedPackagesBox.SelectedItems
-        foreach($package in $selectedPackages)
-        {
-            $unselectedPackagesBox.BeginUpdate()
-            $unselectedPackagesBox.Items.Add($package) | Out-Null
-            $unselectedPackagesBox.EndUpdate()
-        }
-        $numUnselectedLabel.text = "Total: $($unselectedPackagesBox.Items.count)"
-
-        $selectedPackagesBox.BeginUpdate()
-        while ($selectedPackagesBox.SelectedItems.count -gt 0) {
-            $selectedPackagesBox.Items.RemoveAt($selectedPackagesBox.SelectedIndex)
-        }
-        $selectedPackagesBox.EndUpdate()
-        $numSelectedLabel.text = "Total: $($selectedPackagesBox.Items.count)"
-    }
-
-    function Remove-AllPackages {
-        foreach($package in $selectedPackagesBox.Items)
-        {
-            $unselectedPackagesBox.BeginUpdate()
-            $unselectedPackagesBox.Items.Add($package) | Out-Null
-            $unselectedPackagesBox.EndUpdate()
-        }
-        $numUnselectedLabel.text = "Total: $($unselectedPackagesBox.Items.count)"
-
-        $selectedPackagesBox.BeginUpdate()
-        $selectedPackagesBox.Items.Clear()
-        $selectedPackagesBox.EndUpdate()
-        $numSelectedLabel.text = "Total: $($selectedPackagesBox.Items.count)"
-    }
-
+    
     # Gather lists of packages (i.e., available, already installed, to install)
     $excludedPackages = @("flarevm.installer.vm", "common.vm")
     $installedPackages = (Get-InstalledPackages).Name
     $packagesToInstall = $configXml.config.packages.package.name | Where-Object { $installedPackages -notcontains $_ }
-    $allPackages = (Get-InstallablePackages).Name | Where-Object { $packagesToInstall -notcontains $_ -and $installedPackages -notcontains $_ -and $excludedPackages -notcontains $_}
+    #$allPackages = (Get-InstallablePackages).Name | Where-Object { $packagesToInstall -notcontains $_ -and $installedPackages -notcontains $_ -and $excludedPackages -notcontains $_}
     $envs = [ordered]@{}
     $configXml.config.envs.env.ForEach({ $envs[$_.name] = $_.value })
 
-    $form                   = New-Object system.Windows.Forms.Form
-    $form.ClientSize        = New-Object System.Drawing.Point(717,740)
-    $form.text              = "FLARE VM Install Customization"
-    $form.TopMost           = $true
-    $form.MaximizeBox       = $false
-    $form.FormBorderStyle   = 'FixedDialog'
-    $form.StartPosition     = 'CenterScreen'
+    
+    #needed to use Find-Package
+    install-packageprovider nuget -force
+    $packagesByCategory = @{}
+    Write-Host "[+ ]Getting categories and listing packages per category"
+    $packagesByCategory = Get-Packages-Categories
+
+    $formEnv                   = New-Object system.Windows.Forms.Form
+    $formEnv.ClientSize        = New-Object System.Drawing.Point(820,523)
+    $formEnv.text              = "FLARE VM Install Customization"
+    $formEnv.TopMost           = $true
+    $formEnv.MaximizeBox       = $false
+    $formEnv.FormBorderStyle   = 'FixedDialog'
+    $formEnv.StartPosition     = 'CenterScreen'
 
     $envVarGroup            = New-Object system.Windows.Forms.Groupbox
     $envVarGroup.height     = 201
     $envVarGroup.width      = 690
     $envVarGroup.text       = "Environment Variable Customization"
     $envVarGroup.location   = New-Object System.Drawing.Point(15,59)
-
-    $packageGroup           = New-Object system.Windows.Forms.Groupbox
-    $packageGroup.height    = 385
-    $packageGroup.width     = 540
-    $packageGroup.text      = "Package Installation Customization"
-    $packageGroup.location  = New-Object System.Drawing.Point(81,285)
 
     $welcomeLabel           = New-Object system.Windows.Forms.Label
     $welcomeLabel.text      = "Welcome to FLARE VM's custom installer. Please select your options below.`nDefault values will be used if you make no modifications."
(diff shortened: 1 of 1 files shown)


install.ps1
```diff
@@ -493,117 +526,36 @@ if (-not $noGui.IsPresent) {
...
-    $form                   = New-Object system.Windows.Forms.Form
-    $form.ClientSize        = New-Object System.Drawing.Point(717,740)
-    $form.text              = "FLARE VM Install Customization"
-    $form.TopMost           = $true
-    $form.MaximizeBox       = $false
-    $form.FormBorderStyle   = 'FixedDialog'
-    $form.StartPosition     = 'CenterScreen'
+    
+    #needed to use Find-Package
+    install-packageprovider nuget -force
```
Ana06, line 540: Make the installation of nuget consistency (same code structure) with the installation of other dependencies like boxstarter and chocolatey. Concretely: check if it is installed if possible, and print a message that is being installed and installed successfully instead of the output:
```
Name                           Version          Source           Summary
----                           -------          ------           -------
nuget                          2.8.5.208        https://onege... NuGet provider for the OneGet meta-package manager
```

Ana06 (changes requested):
@sara-rn the installer here is broken because https://github.com/mandiant/flare-vm/pull/671 broke the installer. Please rebase after https://github.com/mandiant/flare-vm/pull/675 is merged so that the code works.

The fist UI window (install customization) is missing a next button, so I couldn't test the categories page:
image
Please address this issue so that I can review the new window as well. 

Remember adding labels to PRs (in this case `enhacement` is appropriate) and to add the issue the PR closes to the PR description (this PR should address https://github.com/mandiant/flare-vm/issues/578).

install.ps1
```diff
@@ -461,6 +461,39 @@ if (-not $noGui.IsPresent) {
...
+             (New-Object System.Net.WebClient).DownloadFile($categoriesUrl, $categoriesFile)
+             $categories = Get-Content -Path $categoriesFile
+            } catch {
+             Write-Host "`t[!] Failed to get categories from '$categoriesUrl'"
+             Write-Host "`t[!] $_"
+            }
+       }
+
+       Get-Content -Path $categoriesFile | ForEach-Object {
+         $currentCategory = $_
```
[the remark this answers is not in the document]
sara-rn: No need to download the file `categories.txt`

sara-rn:
@Ana06 there is no need to use `Find-Package`, therefore the installation of nuget or download `categories.txt` is not necessary. At the end of the feed from https://www.myget.org/F/vm-packages/Packages there is a `<link rel="next"` with the URL for the next packages.
The packages are selected and the Continue button adds the selected items to the `config.xml`

install.ps1
```diff
@@ -877,18 +667,181 @@ if (-not $noGui.IsPresent) {
...
+            Invoke-WebRequest -Uri $vmPackagesUrl -OutFile $vmPackagesFile -ErrorAction Stop
+        }
+        catch {
+            Write-Error "`t[!] Failed to get vm-packages from '$vmPackagesUrl. Error: $_"
+            exit
+        }
+
+        # Load the XML content
+        [xml]$vm_packages = Get-Content $vmPackagesFile
+
```
Ana06, line 700: [nitpick]
```suggestion
```

install.ps1
```diff
@@ -877,18 +667,181 @@ if (-not $noGui.IsPresent) {
...
         Write-Host "[+] Cancel pressed, stopping installation..."
         Start-Sleep 3
         exit 1
     }
 
+    ################################################################################
+    ## PACKAGE SELECTION BY CATEGORY
+    ################################################################################
+
+    function Get-Packages-Categories {
```
Ana06, line 683: @sara-rn 
> @Ana06 there is no need to use `Find-Package`, therefore the installation of nuget or download `categories.txt` is not necessary. At the end of the feed from https://www.myget.org/F/vm-packages/Packages there is a `<link rel="next"` with the URL for the next packages.
The packages are selected and the Continue button adds the selected items to the `config.xml`

Seeing how complicated the code to use `https://www.myget.org/F/vm-packages/api/v2/Packages?$filter=IsLatestVersion%20eq%20true` is, I think the solution with `Find-Package` was better: much simpler code and easier to maintain.

Ana06 (changes requested):
@sara-rn I was not able to review the installer UI as the installer does not run correctly:
```
Get-Packages-Categories : The term 'Get-Packages-Categories' is not recognized as the name of a cmdlet, function,
script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is
correct and try again.
At C:\Users\flare\Desktop\install.ps1:491 char:27
+     $packagesByCategory = Get-Packages-Categories
+                           ~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (Get-Packages-Categories:String) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : CommandNotFoundException
```
I think is because you have placed the code of `Get-Packages-Categories` after the use of the function (it needs to be defined before). Please ensure the code works as it otherwise make it the review difficult. 

I also recommend you to add some screenshots when making UI changes.

sara-rn:
> @sara-rn I was not able to review the installer UI as the installer does not run correctly:
> 
> ```
> Get-Packages-Categories : The term 'Get-Packages-Categories' is not recognized as the name of a cmdlet, function,
> script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is
> correct and try again.
> At C:\Users\flare\Desktop\install.ps1:491 char:27
> +     $packagesByCategory = Get-Packages-Categories
> +                           ~~~~~~~~~~~~~~~~~~~~~~~
>     + CategoryInfo          : ObjectNotFound: (Get-Packages-Categories:String) [], ParentContainsErrorRecordException
>     + FullyQualifiedErrorId : CommandNotFoundException
> ```
> 
> I think is because you have placed the code of `Get-Packages-Categories` after the use of the function (it needs to be defined before). Please ensure the code works as it otherwise make it the review difficult.
> 
> I also recommend you to add some screenshots when making UI changes.

I executed it without any issues, but when I switched to a different snapshot with Powershell version 5.1 I had the same error

sara-rn:
Screenshot From 2025-03-18 16-59-44

install.ps1
```diff
@@ -494,116 +480,85 @@ if (-not $noGui.IsPresent) {
...
-
-    function Add-AllPackages {
-        foreach($package in $unselectedPackagesBox.Items)
-        {
-            $selectedPackagesBox.BeginUpdate()
-            $selectedPackagesBox.Items.Add($package) | Out-Null
-            $selectedPackagesBox.EndUpdate()
+        catch {
+            Write-Error "`t[!] Failed to get vm-packages from '$vmPackagesUrl. Error: $_"
+            exit
```
Ana06, line 495: It seems we are downloading files in the installer in several different ways:
- `Invoke-WebRequest`
- `(New-Object net.webclient).DownloadFile()`
- `(New-Object System.Net.WebClient).DownloadFile()`

We also handle errors differently. I think we should introduce a function to download files and handle error consistently (also in the `Get-ConfigFile` and remove the error handling from there):
```PS
function Download-File {
    param (
        [string]$fileSource,
        [string]$fileDestination,
        [switch]$exitOnError
    )
    Write-Host "[+] Downloadingfile from '$fileSource'"
    try {
        Invoke-WebRequest -Uri $fileSource -OutFile $fileDestination
    } catch {
        Write-Host "`t[!] Failed to download '$fileSource'"
        Write-Host "`t[!] $_"
        if ($exitOnError) {
            Start-Sleep 3
            exit 1
        }
    }
)
```

Also, I think `-ErrorAction Stop` is not needed.

Ana06 (changes requested):
Awesome work @sara-rn, this is a nice enhacement!

Please **reduce the size of the first window** to match the size of the contained items:
image

In the second window, I think the font too small and consequently difficult to read. Please **increase the font of both the packages and the titles, making the titles a bit bigger** that the packages. It would also make it easier to read to **bold the package name and remove the ` :`**.
image

I think a bit of padding at the end of the list would be good:
image

I think we should limit the description length in VM-Packages, introducing a linter that checks for a maximum of characters. I have created an issue to track this: https://github.com/mandiant/VM-Packages/issues/1328. @sara-rn you can assume a limit in this PR and use it to cut the description if this makes the displaying nicer in the installer.

install.ps1
```diff
@@ -877,18 +734,192 @@ if (-not $noGui.IsPresent) {
...
         Write-Host "[+] Cancel pressed, stopping installation..."
         Start-Sleep 3
         exit 1
     }
 
+    ################################################################################
+    ## PACKAGE SELECTION BY CATEGORY
+    ################################################################################
+
+    function Install-Selected-Packages
```
Ana06, line 750: Please add a short comment to document what the function does in all the added functions.

install.ps1
```diff
@@ -95,6 +95,26 @@ param (
 $ErrorActionPreference = 'Stop'
 $ProgressPreference = 'SilentlyContinue'
 
+# Function to download files using Invoke-WebRequest and handle errors consistently
+function Download-File {
```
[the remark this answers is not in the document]
sara-rn: I didn't use it in Get-Packages-Categories cause it would only display packages from the last downloaded vm-packages.xml

install.ps1
```diff
@@ -95,6 +95,26 @@ param (
...
+    param (
+        [string]$fileSource,
+        [string]$fileDestination,
+        [switch]$exitOnError
+    )
+    Write-Host "[+] Downloading file from '$fileSource'"
+    try {
+         (New-Object System.Net.WebClient).DownloadFile($fileSource,$fileDestination)
```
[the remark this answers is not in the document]
sara-rn: The reason I changed it instead of Invoke-WebRequest is because it seems to be slower due to the progress bar: https://github.com/actions/runner/issues/1550
I agree we should be consistent using one of the other, however the usage of the function doesn't work in Get-Packages-Categories. Let me know what your thoughts are

install.ps1
sara-rn: I stack to your implementation and therefore left Invoke-WebRequest, the reason is that we don't download large files and it case the connection is slow it displays a progress bar. The usage of System.Net.WebClient is faster though.

install.ps1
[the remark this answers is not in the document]
sara-rn: See my previous comment regarding the usage of Invoke-WebRequest
You were right I didn't invoke Download-File correctly with the switch, it has been fixed

sara-rn:
I implemented the changes except for: It would also make it easier to read to bold the package name and remove the  :.
The only way to implement this would be to create a label for the description, however it is not trivial to know where to draw it exactly because the checkbox size is set to auto. An option would be to create checkboxes with the same size but given the different lengths of packages it would not look nice in my opinion.

install.ps1
```diff
@@ -877,18 +716,191 @@ if (-not $noGui.IsPresent) {
...
+        $Panel_Categories.Controls.Add($labelCategory)
+
+        $NumPackages = 0
+        $verticalPosition2 = $verticalPosition + 20
+		$packages= Get-PackagesByCategory -category $category
+		foreach ($package in $packages)
+		{
+			$NumPackages++
+			$checkBox = New-Object System.Windows.Forms.CheckBox
+			$checkBox.Text = $package.PackageName + " : " + $package.PackageDescription
```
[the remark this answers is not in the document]
Ana06: We have decided not to bold the package name as the implementation is complicated. But what about removing the space before the `:`?

### Current
`apimonitor.vm : API Monitor lets you monitor and control API calls made by applications and services.`
### Proposal
`apimonitor.vm: API Monitor lets you monitor and control API calls made by applications and services.`


### Suggested change
```suggestion
			$checkBox.Text = $package.PackageName + ": " + $package.PackageDescription
```

install.ps1
```diff
@@ -877,18 +726,203 @@ if (-not $noGui.IsPresent) {
...
+
+    $FormCategories                            = New-Object system.Windows.Forms.Form
+    $FormCategories.ClientSize                 = New-Object System.Drawing.Point(1015,800)
+    $FormCategories.text                       = "FLAREVM Package selection"
+    $FormCategories.StartPosition              = 'CenterScreen'
+    $FormCategories.TopMost                    = $true
+
+
+    $categories_label                = New-Object system.Windows.Forms.Label
+    $categories_label.text           = "Select packages to install"
```
Ana06, line 808: I think adding some more details here would be useful. I think the best way to do this is adding an `if` here and rendering one of the following messages depending on if the default or a custom configuration is used:

### Default config (likely most common for users using the UI)
```
Select packages to install. The default configuration (recommended for reverse engineering) is pre-selected.
```

### Custom config 
```
Select packages to install. The provided custom configuration is pre-selected.
```

install.ps1
```diff
@@ -877,18 +716,191 @@ if (-not $noGui.IsPresent) {
...
+    # this doesn't work yet
+    $resetButton                 = New-Object system.Windows.Forms.Button
+    $resetButton.text            = "Reset"
+    $resetButton.AutoSize        = $true
+    $resetButton.location        = New-Object System.Drawing.Point(50,750)
+    $resetButton.Font            = New-Object System.Drawing.Font('Microsoft Sans Serif',10)
+    $resetButton.Add_Click({Set-InitialPackages})#it doesn't work
+
+    $allPackagesButton                 = New-Object system.Windows.Forms.Button
+    $allPackagesButton.text            = "Select All"
```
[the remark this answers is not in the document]
Ana06: If we keep the `Select All` button, I think we should render the following information:

```
Selecting all packages considerable increases installation time and it is not desirable for most use cases.
```
I have no strong prefer about how to render it (a confirmation message when you press select all, a small text below the buttons, a pop-up when hovering the `Select All` button, etc.)

Ana06 (commented):
@sara-rn thanks for adding documentation, I find the code much easy to read now! :bouquet: I really like how this how turned out, it is going to make it much easier for our users to packages to install! :cupid: 

[nitpick] I think reducing the first window a bit more in size (to have the same size on the right and the left sides and to keep the buttons part consistent with the next window, which is more narrow) would look better.

Also, what about adding a `Back` button in the second window? This could be addressed in a different PR too.


image
image

install.ps1
Ana06: This is a good point! Latest PowerShell versions support a new flag for `Invoke-WebRequest` (`-ProgressAction SilentlyContinue`) to handle what you are mentioning. But we can't use it, as we need to support older versions.

I think performance is the most important and my main concern is consistency in our code. Let's **use `(New-Object net.webclient).DownloadFile()` consistently in the code** (replace this and the other uses of download functions).

jstrosch:
Going through the PR now - really nice work! I think this will make customization very smooth. I didn't notice any issues and the installation appears to have kicked off without any problems. Just a few thoughts/observations:

- The package selection screen says "continue", when I clicked it I was expecting a confirmation or summary page, instead it jumped right into the installation. Maybe the button could say something along the lines of "Install" or there is one more page before installing. I wasn't quite ready to start the installation - there could have been text that warned me, but I missed it if so.
- I'm sure this is non-trivial, but would be helpful to see package size/total install size if possible - could be a summary page after selecting packages even :)
- I could see this going one step further and offering a "profile" that pre-selects packages. Say I want an Android profile, it would pre-select all of the common Android tools (in their appropriate categories) and ignore the rest - of course, the package selection screen could provide customization but would allow for users to get a pre-selected baseline easier. Other profiles could be Windows Malware, Office Malware, etc. Just thinking out loud a bit :)  These profiles could serve the non-GUI install, as well as reduce clicks during GUI install.

Again, just some thoughts and no issues to report - can't wait to see this PR go live!

Ana06:
Thanks for taking a look @jstrosch, great feedback!

> The package selection screen says "continue", when I clicked it I was expecting a confirmation or summary page, instead it jumped right into the installation. Maybe the button could say something along the lines of "Install" or there is one more page before installing. I wasn't quite ready to start the installation - there could have been text that warned me, but I missed it if so.

I agree, **`Install` instead of `Continue` in the second UI would be clearer** :+1: @sara-rn can you please address this in this PR?

> I'm sure this is non-trivial, but would be helpful to see package size/total install size if possible - could be a summary page after selecting packages even :)

This is a good idea, but I think **it is not possible** (or at least difficult) to implement. Our packages normally don't contain the tool, just a download link to use to install it. What is useful is to know the installed tool size, but there is no way to know without installing the tool. One thing that **we could easily do is to add this information in the description** for packages like visualstudio.vm and pdbresym.vm, that we know increase the size of the VM significantly (reason why they are not in the default configuration). For example: `Requires Substantial Disk Space` of `Requires X Gb of Disk Space`. We can create an issue for this.


> I could see this going one step further and offering a "profile" that pre-selects packages. Say I want an Android profile, it would pre-select all of the common Android tools (in their appropriate categories) and ignore the rest - of course, the package selection screen could provide customization but would allow for users to get a pre-selected baseline easier. Other profiles could be Windows Malware, Office Malware, etc. Just thinking out loud a bit :) These profiles could serve the non-GUI install, as well as reduce clicks during GUI install.

This is more or less already possible providing a configuration file. We could store more than one configuration file in this repository and display them in the UI (for example in the first UI). **The issue with this is that we have to maintain all the configuration files** adding new tools to several of them (and we already forget to add tools now and we have only one config :sweat:). At the moment we have just one configuration file that contains what FLARE think is required to do malware analysis efficiently without an important size increase ([mostly what FLARE uses internally](https://github.com/mandiant/flare-vm/blob/main/virtualbox/configs/win10_flare-vm.yaml#L10)). This is related to https://github.com/mandiant/flare-vm/issues/668.

install.ps1
```diff
@@ -877,18 +726,203 @@ if (-not $noGui.IsPresent) {
...
+    $categories_label.location       = New-Object System.Drawing.Point(30,20)
+    $categories_label.Font           = New-Object System.Drawing.Font('Microsoft Sans Serif',10)
+
+    $Panel_Categories                = New-Object system.Windows.Forms.Panel
+    $Panel_Categories.height         = 700
+    $Panel_Categories.width          = 970
+    $Panel_Categories.location       = New-Object System.Drawing.Point(30,30)
+    $Panel_Categories.AutoScroll     = $true
+
+    # this doesn't work yet
```
MalwareMechanic, line 821: [nitpick] This worked for me, can remove comment

install.ps1
MalwareMechanic: I like what Ana is suggestion ^

install.ps1
```diff
@@ -877,18 +726,203 @@ if (-not $noGui.IsPresent) {
...
+	$FormCategories.controls.AddRange(@($categories_label,$Panel_Categories,$ContinueButton,$resetButton,$allPackagesButton,$cancelButton,$clearPackagesButton))
+	$FormCategories.Add_Shown({$FormCategories.Activate()})
+	$resultCategories = $FormCategories.ShowDialog()
+	if ($resultCategories -eq [System.Windows.Forms.DialogResult]::OK){
+		Install-Selected-Packages
+	} else {
+		Write-Host "[+] Cancel pressed, stopping installation..."
+		Start-Sleep 3
+		exit 1
+	}
```
MalwareMechanic, line 924: [nitpick] I noticed some tabs used in this file instead of spaces. I'd recommend we pick one and stick with it, though it seems we typically use spaces. We might be able to flag future tab usage with a linter? @Ana06 :eyes:

install.ps1
```diff
@@ -493,117 +493,82 @@ if (-not $noGui.IsPresent) {
...
 
-    function Remove-AllPackages {
-        foreach($package in $selectedPackagesBox.Items)
-        {
-            $unselectedPackagesBox.BeginUpdate()
-            $unselectedPackagesBox.Items.Add($package) | Out-Null
-            $unselectedPackagesBox.EndUpdate()
-        }
-        $numUnselectedLabel.text = "Total: $($unselectedPackagesBox.Items.count)"
+ }
```
MalwareMechanic, line 549: [nitpick] Please fix the scoping indentations so they align better

install.ps1
```diff
@@ -493,117 +493,82 @@ if (-not $noGui.IsPresent) {
...
-    $excludedPackages = @("flarevm.installer.vm", "common.vm")
-    $installedPackages = (Get-InstalledPackages).Name
-    $packagesToInstall = $configXml.config.packages.package.name | Where-Object { $installedPackages -notcontains $_ }
-    $allPackages = (Get-InstallablePackages).Name | Where-Object { $packagesToInstall -notcontains $_ -and $installedPackages -notcontains $_ -and $excludedPackages -notcontains $_}
+    # Gather lists of packages
     $envs = [ordered]@{}
     $configXml.config.envs.env.ForEach({ $envs[$_.name] = $_.value })
+    $packagesByCategory = Get-Packages-Categories
+
+
```
MalwareMechanic, line 556: [nitpick] Remove empty lines

install.ps1
```diff
@@ -612,93 +577,7 @@ if (-not $noGui.IsPresent) {
...
-    $numSelectedLabel.location       = New-Object System.Drawing.Point(288,355)
-    $numSelectedLabel.Font           = New-Object System.Drawing.Font('Microsoft Sans Serif',9)
-
-    $doInstallLabel                  = New-Object system.Windows.Forms.Label
-    $doInstallLabel.text             = "To Install"
-    $doInstallLabel.AutoSize         = $true
-    $doInstallLabel.width            = 25
-    $doInstallLabel.height           = 10
-    $doInstallLabel.location         = New-Object System.Drawing.Point(289,20)
-    $doInstallLabel.Font             = New-Object System.Drawing.Font('Microsoft Sans Serif',10)
```
MalwareMechanic, line 580: [nitpic] Remove empty line

install.ps1
```diff
@@ -612,93 +577,7 @@ if (-not $noGui.IsPresent) {
...
 
-    $resetButton                 = New-Object system.Windows.Forms.Button
-    $resetButton.text            = "Reset"
-    $resetButton.width           = 48
-    $resetButton.height          = 18
-    $resetButton.location        = New-Object System.Drawing.Point(485,358)
-    $resetButton.Font            = New-Object System.Drawing.Font('Microsoft Sans Serif',9)
-    $resetButton.Add_Click({Set-InitialPackages})
```
MalwareMechanic, line 581: [nitpic] Remove empty line

install.ps1
```diff
@@ -799,72 +678,42 @@ if (-not $noGui.IsPresent) {
...
+    $cancelButton.height             = 37
+    $cancelButton.location           = New-Object System.Drawing.Point(587,310)
+    $cancelButton.Font               = New-Object System.Drawing.Font('Microsoft Sans Serif',11)
     $cancelButton.DialogResult = [System.Windows.Forms.DialogResult]::Cancel
 
-    $form.controls.AddRange(@($envVarGroup,$packageGroup,$okButton,$cancelButton,$welcomeLabel))
-    $form.AcceptButton = $okButton
-    $form.CancelButton = $cancelButton
-    $packageGroup.controls.AddRange(@($unselectedPackagesBox,$selectedPackagesBox,$removePackageButton,$removeAllPackageButton,$addPackageButton,$addAllPackageButton,$dontInstallLabel,$doInstallLabel,$numSelectedLabel,$numUnselectedLabel,$resetButton))
-    $envVarGroup.controls.AddRange(@($vmCommonDirText,$vmCommonDirSelect,$vmCommonDirLabel,$toolListDirText,$toolListDirSelect,$toolListDirLabel,$toolListShortCutText,$toolListShortcutSelect,$toolListShortcutLabel,$vmCommonDirNote,$toolListDirNote,$toolListShortcutNote,$rawToolsDirText,$rawToolsDirSelect,$rawToolsDirLabel,$rawToolsDirNote,$metapackageNote1,$metapackageNote2,$metapackageNote3))
```
MalwareMechanic, line 697: [nitpick] Remove empty line

install.ps1
```diff
@@ -799,72 +678,42 @@ if (-not $noGui.IsPresent) {
...
 
-    Set-InitialPackages
+    $formEnv.controls.AddRange(@($envVarGroup,$okButton,$cancelButton,$welcomeLabel))
+    $formEnv.AcceptButton = $okButton
+    $formEnv.CancelButton = $cancelButton
+
+    $envVarGroup.controls.AddRange(@($vmCommonDirText,$vmCommonDirSelect,$vmCommonDirLabel,$toolListDirText,$toolListDirSelect,$toolListDirLabel,   $toolListShortCutText,$toolListShortcutSelect,$toolListShortcutLabel,$vmCommonDirNote,$toolListDirNote,$toolListShortcutNote,$rawToolsDirText,$rawToolsDirSelect,$rawToolsDirLabel,$rawToolsDirNote))
+
```
MalwareMechanic, line 703: [nitpick] Remove empty line

install.ps1
```diff
@@ -877,18 +726,203 @@ if (-not $noGui.IsPresent) {
...
+      # Add selected packages
+      $packages = $configXml.SelectSingleNode('//packages')
+      foreach($package in $selectedPackages) {
+           Write-Host ($package)
+           $newXmlNode = $packages.AppendChild($configXml.CreateElement("package"))
+           $newXmlNode.SetAttribute("name", $package)
+       }
+    }
+
+    # Function that reset the checkboxes to match the config.xml
```
MalwareMechanic, line 762:
```suggestion
    # Function that resets the checkboxes to match the config.xml
```

install.ps1
```diff
@@ -877,18 +726,203 @@ if (-not $noGui.IsPresent) {
...
+            if (($checkbox.Checked) -and ($package -notin $packagesToInstall)){
+                $checkBox.Checked = $false
+            }else{
+              if ((-not $checkbox.Checked ) -and ($package -in $packagesToInstall)){
+                 $checkBox.Checked = $true
+              }
+            }
+        }
+    }
+    # Function that checks all the checkboxes
```
MalwareMechanic, line 775:
```suggestion
    }
    
    # Function that checks all the checkboxes
```

install.ps1
```diff
@@ -877,18 +726,203 @@ if (-not $noGui.IsPresent) {
...
+    $ContinueButton.Font       = New-Object System.Drawing.Font('Microsoft Sans Serif',12)
+
+    $cancelButton            = New-Object system.Windows.Forms.Button
+    $cancelButton.text       = "Cancel"
+    $cancelButton.width      = 97
+    $cancelButton.height     = 37
+    $cancelButton.location   = New-Object System.Drawing.Point(670,750)
+    $cancelButton.Font       = New-Object System.Drawing.Font('Microsoft Sans Serif',12)
+    $cancelButton.DialogResult = [System.Windows.Forms.DialogResult]::Cancel
+
```
MalwareMechanic, line 858: [nitpick] Remove empty line

install.ps1
```diff
@@ -877,18 +726,203 @@ if (-not $noGui.IsPresent) {
...
+    $clearPackagesButton.Font            = New-Object System.Drawing.Font('Microsoft Sans Serif',10)
+    $clearPackagesButton.Add_Click({Clear-AllPackages})
+
+    $ContinueButton            = New-Object system.Windows.Forms.Button
+    $ContinueButton.text       = "Continue"
+    $ContinueButton.width      = 97
+    $ContinueButton.height     = 37
+    $ContinueButton.DialogResult   = [System.Windows.Forms.DialogResult]::OK
+    $ContinueButton.location   = New-Object System.Drawing.Point(570,750)
+    $ContinueButton.Font       = New-Object System.Drawing.Font('Microsoft Sans Serif',12)
```
MalwareMechanic, line 849:
```suggestion
    $installButton            = New-Object system.Windows.Forms.Button
    $installButton.text       = "Install"
    $installButton.width      = 97
    $installButton.height     = 37
    $installButton.DialogResult   = [System.Windows.Forms.DialogResult]::OK
    $installButton.location   = New-Object System.Drawing.Point(570,750)
    $installButton.Font       = New-Object System.Drawing.Font('Microsoft Sans Serif',12)
```

install.ps1
```diff
@@ -877,18 +726,203 @@ if (-not $noGui.IsPresent) {
...
+    $emptyLabel.Width = 20
+    $emptyLabel.Height = 10
+    $emptyLabel.location       = New-Object System.Drawing.Point(10,$posEnd)
+    $Panel_Categories.Controls.Add($emptyLabel)
+
+    # Select packages that are in the config.xml
+    Set-InitialPackages
+
+
+	$FormCategories.controls.AddRange(@($categories_label,$Panel_Categories,$ContinueButton,$resetButton,$allPackagesButton,$cancelButton,$clearPackagesButton))
```
MalwareMechanic, line 915:
```suggestion
	$FormCategories.controls.AddRange(@($categories_label,$Panel_Categories,$installButton,$resetButton,$allPackagesButton,$cancelButton,$clearPackagesButton))
```

install.ps1
```diff
@@ -877,18 +726,203 @@ if (-not $noGui.IsPresent) {
...
+        param (
+         [string]$category
+        )
+        return $packagesByCategory[$category]
+    }
+
+    Add-Type -AssemblyName System.Windows.Forms
+    [System.Windows.Forms.Application]::EnableVisualStyles()
+
+    $FormCategories                            = New-Object system.Windows.Forms.Form
```
MalwareMechanic, line 800: [nitpick] Let's stick with came-case with the first letter lowercase, like `$formCategories`

install.ps1
```diff
@@ -877,18 +726,203 @@ if (-not $noGui.IsPresent) {
...
+    $FormCategories.ClientSize                 = New-Object System.Drawing.Point(1015,800)
+    $FormCategories.text                       = "FLAREVM Package selection"
+    $FormCategories.StartPosition              = 'CenterScreen'
+    $FormCategories.TopMost                    = $true
+
+
+    $categories_label                = New-Object system.Windows.Forms.Label
```
MalwareMechanic, line 807: [nitpick] Let's stick with came-case with the first letter lowercase, like `$categoriesLabel`

install.ps1
```diff
@@ -877,18 +726,203 @@ if (-not $noGui.IsPresent) {
...
+    $categories_label.text           = "Select packages to install"
+    $categories_label.AutoSize       = $true
+    $categories_label.width          = 25
+    $categories_label.height         = 10
+    $categories_label.location       = New-Object System.Drawing.Point(30,20)
+    $categories_label.Font           = New-Object System.Drawing.Font('Microsoft Sans Serif',10)
+
+    $Panel_Categories                = New-Object system.Windows.Forms.Panel
```
MalwareMechanic, line 815: [nitpick] Let's stick with came-case with the first letter lowercase, like `$panelCategories`

MalwareMechanic (changes requested):
Mostly nitpicks here, PR overall looks great :smiley: 

Please check that the "Cancel" buttons are working as intended. I ran the script from an elevated PowerShell console and when I cancelled the GUI disappeared, but I was stuck at PowerShell console. I didn't see any exit messages and `ctrl+c` didn't work to exit it. However, when I debugged it in PowerShell ISE it seemed to work fine and also print the exit message. :shrug:

sara-rn:
> Mostly nitpicks here, PR overall looks great 😃
> 
> Please check that the "Cancel" buttons are working as intended. I ran the script from an elevated PowerShell console and when I cancelled the GUI disappeared, but I was stuck at the PowerShell console. I didn't see any exit messages and `ctrl+c` didn't work to exit it. However, when I debugged it in PowerShell ISE it seemed to work fine and also print the exit message. 🤷

@MalwareMechanic I cannot reproduce the scenario you mention, both cancel buttons are expected to exit and print the message. I am not sure how to address this.

sara-rn:
Thanks @Ana06 @MalwareMechanic and @jstrosch for the feedback, much appreciated! I have addressed the proposed changes. I believe the overall look and feel can be improved such as the usage of icons or the background.pn that we don't use. @jstrosch I like your idea regarding the profile, this is what commando-vm has, but I agree with Ana we would need to maintain it, no issues if everyone finds it useful. An idea for another PR would be to save the selected packages into a custom config for example so next time you can provide it as a command-line. Just some random thoughts. Thanks again all!!

sara-rn:
> @sara-rn thanks for adding documentation, I find the code much easy to read now! 💐 I really like how this how turned out, it is going to make it much easier for our users to packages to install! 💘
> 
> [nitpick] I think reducing the first window a bit more in size (to have the same size on the right and the left sides and to keep the buttons part consistent with the next window, which is more narrow) would look better.
> 
> Also, what about adding a `Back` button in the second window? This could be addressed in a different PR too.
> 
> image image

I reduced the height of the first Window, not sure if I should also move the Install and Cancel buttons to the right in the second Window or its already fine.

jstrosch (approved):

install.ps1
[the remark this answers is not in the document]
Ana06: I get an expection when pressing the button:
image

install.ps1
```diff
@@ -877,12 +718,200 @@ if (-not $noGui.IsPresent) {
...
+    $labelCategories.Font           = New-Object System.Drawing.Font('Microsoft Sans Serif',10)
+
+    $panelCategories                = New-Object system.Windows.Forms.Panel
+    $panelCategories.height         = 700
+    $panelCategories.width          = 970
+    $panelCategories.location       = New-Object System.Drawing.Point(30,30)
+    $panelCategories.AutoScroll     = $true
+
+    $resetButton                 = New-Object system.Windows.Forms.Button
+    $resetButton.text            = "Reset"
```
Ana06, line 817: [nitpick] I think renaming this button to `Rest to default config` or `Reset to custom config` (depending if a custom config is provided) would make it easier to understand what _reset_ means.

Ana06 (changes requested):

Ana06:
@MalwareMechanic 
> Please check that the "Cancel" buttons are working as intended. I ran the script from an elevated PowerShell console and when I cancelled the GUI disappeared, but I was stuck at the PowerShell console. I didn't see any exit messages and ctrl+c didn't work to exit it. However, when I debugged it in PowerShell ISE it seemed to work fine and also print the exit message

I think the exit message not being rendered is caused by https://github.com/mandiant/flare-vm/issues/541 (confusing but unrelated to the changes here).

I can't reproduce the stuck after exist on the PowerShell console, but the way to exit hasn't been changed in this PR. So if there is an issue, I think it is not caused by these changes. @MalwareMechanic can you double-check that this happens consistently and open an issue if so? Then we can investigate the issue after this has been merged.

Ana06:
@sara-rn
> I reduced the height of the first Window, not sure if I should also move the Install and Cancel buttons to the right in the second Window or its already fine.

I think it is better to keep the windows as consistent and similar as possible.

install.ps1
sara-rn: I believe the size of the reset would be too large, not sure if that will look good. How about adding it to the label, for example: Select packages to install. The provided custom configuration is pre-selected. Click on the reset button to restore the configuration.

Ana06: This is a good idea! :+1:

Ana06 (approved):
@sara-rn I have tested again and everything looks as expected. This is a great improvement, thanks for all the work! :rocket: 

[nitpick] I think it would be nice to add a Back button in the second window (to go back to the first one).

sara-rn:
@MalwareMechanic since you proposed changes it seems you need to approve it for the Pr to be merged, can you take another quick look? I addressed all the changes you proposed but I force-pushed it after that so it seems there are pending changes.

PR #685, opened 2025-04-24
[config] Add new packages

Ana06:
Add recently added packages to the default configuration.

@mandiant/flare-vm any other package that should be added?


sara-rn (approved):
looks good to me

PR #686, opened 2025-04-25
[config] Add psnotify

Ana06:
@mike-hunhoff proposed to add psnotify (in combination with GarbageMan) in https://github.com/mandiant/VM-Packages/issues/247#issuecomment-1561464792. We created a package for both some time ago, but forgot to add psnotify to the default configuration.


sara-rn (approved):
LGTM

PR #694, opened 2025-05-08
Enhance FLARE-VM Installer

### LINKED ISSUE #669: Handle community packages better
### Details

The current config include the community!package dotnet3.5. I think we do not display this package in the installer UI or allow to add other community packages. We could have some thing like a text box for these packages (for example comma separated list) and a validation similar to the one in commando VM:

Image

### LINKED ISSUE #692: How to handle Commando VM categories in the installer?
### Details

At the moment we order the categories in the installer alphabetically. This has the issue that some categories that are mainly for Commando VM (and likely never installed in a FLARE-VM installation) like `Command and Control`, `Credential Access`, `Explotation` and `Forensic` appear at the top. I propose to remove the Commando VM categories completely from the installer, by keeping a list of hardcoded categories in the installer. The Commando VM packages could still be installed in the same way as community packages: https://github.com/mandiant/flare-vm/issues/669. We should also add a link to https://github.com/mandiant/VM-Packages/wiki/Packages in the installer (next to the textbox) to make it easier for user to look for packages.

@mandiant/vms opinions?

sara-rn:
As part of the default config there are chocolatey packages that were not displayed previously.
Only display packages sorted by category that are relevant to RE, excluded hardcoded categories.
Add a listbox that displays additional packages that are either chocolatey packages or belong to excluded categories typically used in Commando-VM. 
Add a new functionality to search across chocolatey packages including FLARE-VM packages to add new packages
Closes #669  
Closes #692


install.ps1
```diff
@@ -766,25 +797,89 @@ if (-not $noGui.IsPresent) {
...
+        $additionalPackagesBox.BeginUpdate()
+        while ($additionalPackagesBox.SelectedItems.count -gt 0) {
+            $additionalPackagesBox.Items.RemoveAt($additionalPackagesBox.SelectedIndex)
+        }
+        $additionalPackagesBox.EndUpdate()
     }
 
     Add-Type -AssemblyName System.Windows.Forms
     [System.Windows.Forms.Application]::EnableVisualStyles()
+    $excludedCategories=@('Command and Control','Credential Access','Exploitation','Forensic','Lateral Movement', 'Payload Development','Privilege Scalation','Reconnaissance','Wordlists')
```
Ana06, line 874: I think we need to exclude `Web Application` as well. :thinking:

Ana06: [discussion] I am not sure if it is better to have a list of excluded categories or a list of categories. We have to main the list in either case. @mandiant/flare-vm opinions?

install.ps1
```diff
@@ -903,6 +998,147 @@ if (-not $noGui.IsPresent) {
...
+                $addPackageButton.enabled = $true
+        } else {
+                $chocoPackage = Get-ChocoPackage -PackageName $packageTextBox.Text
+            if ($chocoPackage) {
+            $chocoPackageErrorLabel.text = "Found Choco package"
+            $chocoPackageErrorLabel.ForeColor = $successColor
+            $chocoPackageErrorLabel.Visible = $true
+            $addPackageButton.enabled = $true
+            }else {
+                $chocoPackageErrorLabel.text = "package not found"
```
Ana06, line 1112: Use consistent capitalization
```suggestion
                $chocoPackageErrorLabel.text = "Package not found"
```

install.ps1
```diff
@@ -914,6 +1150,9 @@ if (-not $noGui.IsPresent) {
         exit 1
     }
 
+
+
+
```
Ana06, line 1155: Unrelated change
```suggestion
```

Ana06 (changes requested):
@sara-rn this is going in the right direction, it is a nice enhancement! :rocket: 

Please **remember to add screenshots to the PR description when doing visual changes**, like changes to the UI, and to add a comment with a screenshot every time you update the PR significantly.  Ensure you **update [the installer screenshot in the README](https://github.com/mandiant/flare-vm?tab=readme-ov-file#installer-gui)** (after we are done discussing the UI) in this PR. We forgot this in the previous PR. 

image

We need to do some improvements to make the new UI nicer and easier to read, using better the space (at the moment there is a lot space to the right of the textbox and some things are placed a bit weird):
- **Increase the font**. Font smaller than the currently used for the package description is difficult to read.
- **Bold and increase the font of both the text on the top (`Select packages to install`) and `Additional packages to install`**. They are titles and should be a bit more visible.
- **Move the `Add package` button next to the `Find Package`**.
- **Split the explanation in longer lines and use bullet points, using the space on the right**:
```
FLARE-VM uses Chocolatey packages. You can add additional packages from:
- Community packages: https://community.chocolatey.org/packages
- VM-Packages: https://github.com/mandiant/VM-Packages/wiki/Packages
```
- **Make the URLs clickable**. If this is not possible, we need a way to short at least the VM-Packages URL, which is very long.
- **Add a bit of space** after the package list by category and before the `[Reset] [Select All][Clear] [Buttons]`. It looks a bit too compact. If needed you can make the window a bit taller to achieve this.
- **Correct typo in Windows title**: Change `FLAREVM package selection` by `FLARE-VM package section`.

Looking for packages in the community feed seems to take some time. What about **adding a message `Finding package...`** when pressing the button after the package is found/not found to provide feedback to the user? We should add this also for the VM-Packages for consistency, although they seem to be found much faster.

If after finding the package, you modify the package name and then press `Add Package` a package that hasn't been found is added. **The `Add Package` button should be disabled if the package name in the text box is edited**. An example with `microsoft-edge`:
image

I find the error when I add `magika.vm` confusing. I think we should **allow to also add packages in the list of package with category via the search textbox**. Functionality wise it doesn't matter if they end up twice in the config file.
image

Loading the second UI window (the package selection) takes some time. The first window is closed and the user doesn't know what is going on. I think we should **initialize `$packagesByCategory` (doing the querying of packages per category) before opening the first installer window.** This way the second window will open much faster. This should be easy to implement, just moving the call to `Get-Packages-Categories`.

[nitpick] It is easy to copy trailing whitespaces when copying package names. What about **removing trailing whitespaces at the beginning/end of the package name**, to ensure it is found and added? If you don't implement this as part of this PR, please create an issue (before the PR is merged to avoid forgetting it).
image

[nitpick] **It shouldn't be possible to add a package twice**. If you don't implement this as part of this PR, please create an issue (before the PR is merged to avoid forgetting it). An example with `dumpert.vm`:
image

sara-rn:
@Ana06 all the changes have been addressed in a separate commit

install.ps1
```diff
@@ -469,6 +469,10 @@ Start-Sleep 1
 $configXml = [xml](Get-Content $configPath)
 
 if (-not $noGui.IsPresent) {
+
```
Ana06, line 472: [nitpick]
```suggestion
```

install.ps1
```diff
@@ -766,70 +818,140 @@ if (-not $noGui.IsPresent) {
...
+    $formCategories.text                       = "FLARE-VM Package selection"
     $formCategories.StartPosition              = 'CenterScreen'
     $formCategories.TopMost                    = $true
 
     if ([string]::IsNullOrEmpty($customConfig)) {
-        $textLabel = "Select packages to install. The default configuration (recommended for reverse engineering) is pre-selected.`nClick on the reset button to restore the default configuration."
+        $textLabel = "The default configuration (recommended for reverse engineering) is pre-selected.`nClick on the reset button to restore the default configuration."
     } else {
-	    $textLabel = "Select packages to install. The provided custom configuration is pre-selected.`nClick on the reset button to restore the custom configuration."
+	    $textLabel = "The provided custom configuration is pre-selected.`nClick on the reset button to restore the custom configuration."
```
Ana06, line 905: I think we should remove the new line now that the message is shorter (as `Select packages to install` is now in a different line) and display everything in the same line.

Ana06 (changes requested):
@sara-rn thanks for making the improvements!

As I said in my previous review, please provide a screenshot in a comment when making changes related to the UI:
image

It may not be clear for the user that `Community Packages` and `FLARE-VM Packages` are clickable. I think we should **display the URL in addition to the name and make only the URL clickable** (as I proposed in my previous review):
```
FLARE-VM uses Chocolatey packages. You can add additional packages from:
- Community packages: https://community.chocolatey.org/packages
- VM-Packages: https://github.com/mandiant/VM-Packages/wiki/Packages
```

I think we should **remove the `:` in `Additional packages to install:`**, as the other titles do not have it. I think it would also be easier to understand if we **add a bit more of space before `Additional packages to install:`** and **move the text `FLARE-VM uses Chocolatey packages...` a bit down** so that it is not at the same level than the title `Additional packages to install`.

**Please add more details to the error for a duplicated package,** for example: `Error to add the package: duplicated`.
image


[nitpick] **Trailing whitespaces are still not removed**. If you don't plan to implement it, please open an issue to track the improvement suggestion.
image


**Ensure the linter passes and squash the commits** (we do not keep `fixup` commits in order to maintain a clean commit history). Please also **update [the installer screenshot in the README](https://github.com/mandiant/flare-vm?tab=readme-ov-file#installer-gui)**, I don't expect the UI to change more after you address the proposed changes.

sara-rn:
Addressed changes.
I changes slightly the message below `Select Packages to install` because it was a bit too long. Let me know.
Screenshot From 2025-05-30 19-10-26

Ana06 (changes requested):
The installer works perfectly, this is a nice improvement. Thanks @sara-rn for all the work and for addressing the feedback!

**It is still missing updating the screenshot of the installer in the README.** @sara-rn do you have time to add it? Otherwise we can merge the PR as it is and I'll add it in a separate PR.

image

Ana06 (changes requested):
@sara-rn how have you taken the screenshot? They are very small and difficult to read in the README: https://github.com/mandiant/flare-vm/blob/df4d96ba84b203f275c86b8657fbb9e2574f0e48/README.md

Ana06 (approved):
Thanks @sara-rn! :rocket:

PR #700, opened 2025-06-24
Fix typo in the excluded categories

sara-rn:
Fix a typo in the list of hardcoded categories to exclude the Privilege Escalation category


Ana06 (approved):
Thanks @sara-rn! :+1:

PR #707, opened 2025-07-15
Add gui preinstall checks

sara-rn:
Based on the code from Commando-VM, created a new Window that performs pre-installation checks. The same pre-checks that are done if the `noGui` is provided.


Ana06 (changes requested):
@sara-rn thanks for working on this enhacements!

As already mentioned in your previous PRs: **when making UI changes, add a screenshot to make the review of the PR easier.**
image



**This code is duplicating a lot of code and logic in both the UI and command line**. This makes it difficult to maintain the code and keep the logic consistent. As said in , we have to ensure we do not duplicate code to make it easy to keep both views consistent. One way to implement this is to have a **checks array** with an entry per check with the following information:
`(name, description, function, allow_continue)`

For example:
`("PowerShell version >= 5", Check-PowerShell-Version, False)`

The function should return an info message that we can display in the command line (as we do at the moment) and in the UI (extra info when a check fails, as the text below the check title). I think we can summarize some of the info messages if needed. For example:

```PS
function Check-PowerShell-Version {
    $psVersion = $PSVersionTable.PSVersion
    if ($psVersion -lt [System.Version]"5.0.0") {
         return "Your PowerShell version ($psVersion) is not supported"
    }
}
```

**The `PSAvoidUsingWMICmdlet` linter complain should be easy to fix by replacing the used functions as explained in https://learn.microsoft.com/en-us/powershell/utility-modules/psscriptanalyzer/rules/avoidusingwmicmdlet?view=ps-modules** In any case, focus on implementing the code using a checks array, as I think this may fix the other offense as well.

sara-rn:
> @sara-rn thanks for working on this enhacements!
> 
> As already mentioned in your previous PRs: **when making UI changes, add a screenshot to make the review of the PR easier.** image
> 
> **This code is duplicating a lot of code and logic in both the UI and command line**. This makes it difficult to maintain the code and keep the logic consistent. As said in , we have to ensure we do not duplicate code to make it easy to keep both views consistent. One way to implement this is to have a **checks array** with an entry per check with the following information: `(name, description, function, allow_continue)`
> 
> For example: `("PowerShell version >= 5", Check-PowerShell-Version, False)`
> 
> The function should return an info message that we can display in the command line (as we do at the moment) and in the UI (extra info when a check fails, as the text below the check title). I think we can summarize some of the info messages if needed. For example:
> 
> ```postscript
> function Check-PowerShell-Version {
>     $psVersion = $PSVersionTable.PSVersion
>     if ($psVersion -lt [System.Version]"5.0.0") {
>          return "Your PowerShell version ($psVersion) is not supported"
>     }
> }
> ```
> 
> **The `PSAvoidUsingWMICmdlet` linter complain should be easy to fix by replacing the used functions as explained in https://learn.microsoft.com/en-us/powershell/utility-modules/psscriptanalyzer/rules/avoidusingwmicmdlet?view=ps-modules** In any case, focus on implementing the code using a checks array, as I think this may fix the other offense as well.

in the proposed function you would return a message,then would you check if a `Test-Powershell-Version` (the linter doesn't allow a function that starts with `Check-`) is not succesful? by check if the return value is not equal to "" ??

Ana06:
@sara-rn 

> in the proposed function you would return a message,then how would you check if a Test-Powershell-Version (the linter doesn't allow a function that starts with Check-) is not successful? by checking if the return value is not equal to "" ?? 

An empty string is the same as False for PowerShell:
```PS
$error_info = Test-PowerShell-Version
if ($error_info) {
  // The check has failed
}
```

> you still want to display messages via Write-Host even if the GUI is enabled right? Following the Commando-vm example I didn't display any mesage in the console when the gui is allowed (only when the Test-Webconnection was invoked cause we invoked three times and we should display at least there are some checks going on (also we really need to check connectivity of the three domains? but that's a different issue). In my opinion this might be confusing for the user cause you need to look at both the window and the console.

No. My proposal was to display the same message in the console and the GUI (only in one of them), but to display the exact same message by returning it by the Test function if the check fails.

Regarding the internet check, you can perform the check before displaying the UI. This is similar to preloading the packages for next UI view. I think we do need to check the three domains to avoid issues, but we do not need to log them, just a message for failure is enough.

Ana06:
> @Ana06 how about the Hint messages we display? Shall we only display them in the console right?

No, my proposal was to display the exact same messages in both the console and the UI, but shortening them a bit if needed. For example: `Re-run the script after updating the execution policy with "Set-ExecutionPolicy Unrestricted"`. I think some of these messages are too long and I think keeping both views consistent is important.

sara-rn:
> Re-run the script after updating the execution policy with "Set-ExecutionPolicy Unrestricted"

Thanks @Ana06 for the clarification, then all these links to articles would disappear as you would only summarize it in a single line:
```
			Write-Host "`t[!] A minimum of 60 GB hard drive space is preferred. Please increase hard drive space of the VM, reboot, and retry install" -ForegroundColor Red
			Write-Host "`t[+] If you have multiple drives, you may change the tool installation location via the envrionment variable %RAW_TOOLS_DIR% in config.xml or GUI" -ForegroundColor Yellow
			Write-Host "`t[+] However, packages provided from the Chocolatey community repository will install to their default location" -ForegroundColor Yellow
			Write-Host "`t[+] See: https://stackoverflow.com/questions/19752533/how-do-i-set-chocolatey-to-install-applications-onto-another-drive" -ForegroundColor Yellow
```
or these knowledge articles:
```
Write-Host "`t[+] Hint: https://support.microsoft.com/en-us/windows/prevent-changes-to-security-settings-with-tamper-protection-31d51aaa-645d-408e-6ce7-8d7f8e593f87" -ForegroundColor Yellow
			Write-Host "`t[+] Hint: https://www.tenforums.com/tutorials/123792-turn-off-tamper-protection-windows-defender-antivirus.html" -ForegroundColor Yellow
			Write-Host "`t[+] Hint: https://github.com/jeremybeaume/tools/blob/master/disable-defender.ps1" -ForegroundColor Yellow
			Write-Host "`t[+] Hint: https://lazyadmin.nl/win-11/turn-off-windows-defender-windows-11-permanently/" -ForegroundColor Yellow
			Write-Host "`t[!] Please disable Windows Defender through Group Policy, reboot, and rerun installer" -ForegroundColor Red
			Write-Host "`t[+] Hint: https://stackoverflow.com/questions/62174426/how-to-permanently-disable-windows-defender-real-time-protection-with-gpo" -ForegroundColor Yellow
			Write-Host "`t[+] Hint: https://www.windowscentral.com/how-permanently-disable-windows-defender-windows-10" -ForegroundColor Yellow
			
```

sara-rn:
Added a message when the installation cannot continue in both cli and gui:
Screenshot From 2025-07-19 13-39-55

sara-rn:
> @sara-rn thanks for working on this enhacements!
> 

> **This code is duplicating a lot of code and logic in both the UI and command line**. This makes it difficult to maintain the code and keep the logic consistent. As said in , we have to ensure we do not duplicate code to make it easy to keep both views consistent. One way to implement this is to have a **checks array** with an entry per check with the following information: `(name, description, function, allow_continue)`
> 
> For example: `("PowerShell version >= 5", Check-PowerShell-Version, False)`
> 
> The function should return an info message that we can display in the command line (as we do at the moment) and in the UI (extra info when a check fails, as the text below the check title). I think we can summarize some of the info messages if needed. For example:
> 
> ```postscript
> function Check-PowerShell-Version {
>     $psVersion = $PSVersionTable.PSVersion
>     if ($psVersion -lt [System.Version]"5.0.0") {
>          return "Your PowerShell version ($psVersion) is not supported"
>     }
> }
> ```
> 
I implemented all the Check functions as you suggested, however I didn't implement the checks_array because it requires a lot of duplication in any case as there are three components in the GUI that are created per check and this cannot be done dynamically. If a check is not successful we have to manually show the $error_info in the label and change the color from grey to red, so it not only shows False but the error message gets updated.

install.ps1
```diff
@@ -479,96 +386,488 @@ if (-not $noGui.IsPresent) {
...
+		$WindowsDefenderLabel           = New-Object system.Windows.Forms.Label
+		$WindowsDefenderLabel.text      = "Windows Defender Disabled"
+		$WindowsDefenderLabel.AutoSize  = $true
+		$WindowsDefenderLabel.width     = 25
+		$WindowsDefenderLabel.height    = 10
+		$WindowsDefenderLabel.location  = New-Object System.Drawing.Point(15,104)
+		$WindowsDefenderLabel.Font      = New-Object System.Drawing.Font('Microsoft Sans Serif',12,[System.Drawing.FontStyle]([System.Drawing.FontStyle]::Bold))
+
+		$WindowsReleaseLabel             = New-Object system.Windows.Forms.Label
+		$WindowsReleaseLabel.text        = "Compatible Windows Release"
```
Ana06, line 441:
```suggestion
		$WindowsReleaseLabel.text        = "Tested Windows Version"
```

install.ps1
```diff
@@ -479,96 +386,488 @@ if (-not $noGui.IsPresent) {
...
+		$RunningVMTooltip.text           = "Only run this script inside a Virtual Machine"
+		$RunningVMTooltip.AutoSize       = $true
+		$RunningVMTooltip.width          = 25
+		$RunningVMTooltip.height         = 10
+		$RunningVMTooltip.location       = New-Object System.Drawing.Point(15,219)
+		$RunningVMTooltip.Font           = New-Object System.Drawing.Font('Microsoft Sans Serif',10)
+		$RunningVMTooltip.ForeColor      = $grayedColor
+
+		$WindowsReleaseTooltip           = New-Object system.Windows.Forms.Label
+		$WindowsReleaseTooltip.text      = "Ensure your Windows version is supported"
```
Ana06, line 595:
```suggestion
		$WindowsReleaseTooltip.text      = "You are likely to run into issue when using a non tested version"
```

Ana06 (commented):
- Change the order of the checks in the UI to match the console
- Change color of non mandatory checks to orange (instead of red) when they fail in hte Ui and in the console (`ForegorundColor`)
- Make the grey color slightly darker
- Make the window a bit wider so that we have more space for the checks (to match next view)

Ana06 (changes requested):
@sara-rn the UI looks good to me now, thanks for the changes! But **these changes break the installer in my Windows 11 VM**, so we can't merge the PR as it will break the FLARE-VM installer for some users:
image

The previous version of the installer was working correctly:
image

Appart from that I have noticed that the individual internet connectivity checks are still displayed to the console. Should we create an issue to remove them in a separate PR? Or do we think it is ok to display them?
image

Ana06:
@sara-rn please rebase to solve the merge conflicts

sara-rn:
> Appart from that I have noticed that the individual internet connectivity checks are still displayed to the console. Should we create an issue to remove them in a separate PR? Or do we think it is ok to display them? image
@Ana06 that's on purpose, I left it because it is quite slow to check connectivity of the three domains so it makes sense to  display the user even in GUI that that the three check are taking place instead of not displaying anything.

sara-rn:
> @sara-rn the UI looks good to me now, thanks for the changes! But **these changes break the installer in my Windows 11 VM**, so we can't merge the PR as it will break the FLARE-VM installer for some users: image
> 
> The previous version of the installer was working correctly: image
> 
> Appart from that I have noticed that the individual internet connectivity checks are still displayed to the console. Should we create an issue to remove them in a separate PR? Or do we think it is ok to display them? image

@Ana06 It should work on Windows 11 but there seems to be a problem with this Windows Build. It is equivalent to using the cmdlet `Get-MpComputerStatus` which also fails with the error `Get-MpComputerStatus : Provider load failure`
so we have the option to simply run it between try / catch and therefore this check wouldn't work or revert the implementation to what we had before that works well in this Windows 11 build which would be: ```
function Test-DefenderAndTamperProtection {
	try {
        $tpEnabled = Get-ItemProperty -Path "HKLM:\SOFTWARE\Microsoft\Windows Defender\Features" -Name "TamperProtection" -ErrorAction Stop
        if ($tpEnabled.TamperProtection -eq 5) {
            return "Disable Tamper Protection, reboot, and rerun installer"
        } 
    } catch {
		return "Tamper Protection is either not enabled or not detected"
	}
	$defender = Get-Service -Name WinDefend -ea 0
	if ($null -ne $defender) {
        if ($defender.Status -eq "Running") {
			 return "Disable Windows Defender through Group Policy, reboot, and rerun installer"
	    }
	}
}
```

**Comment by sara-rn** [comment]
This is the error I get with the old implementation when I ran it in my Windows 10 VM after removing the try/catch:
```
Get-ItemProperty : Cannot find path 'HKLM:\SOFTWARE\Microsoft\Windows Defender\Features' because it does not exist.
At C:\Users\flaretest\Desktop\install_cat.ps1:192 char:22
+ ... tpEnabled = Get-ItemProperty -Path "HKLM:\SOFTWARE\Microsoft\Windows  ...
```

File: install.ps1 [hunk 3]
```diff
@@ -479,102 +425,489 @@ if (-not $noGui.IsPresent) {
...
+		$WindowsDefender.location = New-Object System.Drawing.Point(24,409)
+		$WindowsDefender.Font = New-Object System.Drawing.Font('Microsoft Sans Serif',12,[System.Drawing.FontStyle]([System.Drawing.FontStyle]::Bold))
+		$WindowsDefender.ForeColor = $orangeColor
+
+		################################# Check Tooltip Controls #################################
+		$verticalPosition = 41
+
+		# $PSVersionTooltip
+		$PSVersionTooltip = New-Object system.Windows.Forms.Label
+		$PSVersionTooltip.text = "Minimum Powershell version 5 (mandatory)"
```
Ana06, line 627: [nitpick]
```suggestion
		$PSVersionTooltip.text = "Powershell version must be >= 5 (mandatory)"
```

install.ps1
```diff
@@ -479,102 +425,489 @@ if (-not $noGui.IsPresent) {
...
+		$PSVersionTooltip.AutoSize = $true
+		$PSVersionTooltip.location = New-Object System.Drawing.Point(15,$verticalPosition)
+		$PSVersionTooltip.Font = New-Object System.Drawing.Font('Microsoft Sans Serif',10)
+		$PSVersionTooltip.ForeColor = $grayedColor
+		$verticalPosition += 44
+
+		# $RunningAsAdminTooltip
+		$RunningAsAdminTooltip = New-Object system.Windows.Forms.Label
+		$RunningAsAdminTooltip.text = "Please run this script as Administrator (mandatory)"
```
Ana06, line 636: [nitpick]
```suggestion
		$RunningAsAdminTooltip.text = "You must run the script as Administrator (mandatory)"
```

install.ps1
```diff
@@ -169,305 +136,284 @@ function Get-ConfigFile {
...
+	}
+}
 
-    # Check if Tamper Protection is disabled
-    Write-Host "[+] Checking if Windows Defender Tamper Protection is disabled..."
+function Test-Admin {
+    try {
+		$currentPrincipal = New-Object Security.Principal.WindowsPrincipal([Security.Principal.WindowsIdentity]::GetCurrent())
+		if (-not ($currentPrincipal.IsInRole([Security.Principal.WindowsBuiltInRole]::Administrator))){
+			return "Your user is not running as Administrator"
```
Ana06, line 187: [nitpick] To keep all Admin messages consistent
```suggestion
			return "The script is not running as Administrator"
```

install.ps1
```diff
@@ -169,305 +136,284 @@ function Get-ConfigFile {
...
+		}
+	} catch {
+		return "Unable to determine is user is running as Administrator"
```
Ana06, line 190: [nitpick] To keep all Admin messages consistent

```suggestion
		return "Unable to determine if the script is running as Administrator"
```

install.ps1
```diff
@@ -479,102 +425,489 @@ if (-not $noGui.IsPresent) {
...
+		$RunningAsAdminTooltip.text = "Please run this script as Administrator (mandatory)"
+		$RunningAsAdminTooltip.AutoSize = $true
+		$RunningAsAdminTooltip.location = New-Object System.Drawing.Point(15,$verticalPosition)
+		$RunningAsAdminTooltip.Font = New-Object System.Drawing.Font('Microsoft Sans Serif',10)
+		$RunningAsAdminTooltip.ForeColor = $grayedColor
+		$verticalPosition += 44
+
+		# $ExecutionPolicyTooltip
+		$ExecutionPolicyTooltip = New-Object system.Windows.Forms.Label
+		$ExecutionPolicyTooltip.text = "PowerShell: Set-ExecutionPolicy Unrestricted (mandatory)"
```
Ana06, line 645: [nitpick]
```suggestion
		$ExecutionPolicyTooltip.text = "You must enable script execution (mandatory)"
```

install.ps1
```diff
@@ -479,102 +425,489 @@ if (-not $noGui.IsPresent) {
...
+		$WindowsReleaseTooltip.text = "You might run into issues when using a non tested version"
+		$WindowsReleaseTooltip.AutoSize = $true
+		$WindowsReleaseTooltip.location = New-Object System.Drawing.Point(15,$verticalPosition)
+		$WindowsReleaseTooltip.Font = New-Object System.Drawing.Font('Microsoft Sans Serif',10)
+		$WindowsReleaseTooltip.ForeColor = $grayedColor
+		$verticalPosition += 44
+
+		# $RunningVMTooltip
+		$RunningVMTooltip = New-Object system.Windows.Forms.Label
+		$RunningVMTooltip.text = "Only run this script inside a Virtual Machine"
```
Ana06, line 672: [nitpick]
```suggestion
		$RunningVMTooltip.text = "Only run this script inside a Virtual Machine (VM)"
```

install.ps1
```diff
@@ -479,102 +425,489 @@ if (-not $noGui.IsPresent) {
...
+		$usernameContainsSpacesToolTip.text = "Username cannot contain spaces (mandatory)"
+		$usernameContainsSpacesToolTip.AutoSize = $true
+		$usernameContainsSpacesToolTip.location = New-Object System.Drawing.Point(15,$verticalPosition)
+		$usernameContainsSpacesToolTip.Font = New-Object System.Drawing.Font('Microsoft Sans Serif',10)
+		$usernameContainsSpacesToolTip.ForeColor = $grayedColor
+		$verticalPosition += 44
+
+		# $EnoughHardStorageTooltip
+		$EnoughHardStorageTooltip = New-Object system.Windows.Forms.Label
+		$EnoughHardStorageTooltip.text = "Have at least 60GB of available storage"
```
Ana06, line 690:
```suggestion
		$EnoughHardStorageTooltip.text = "A minimum of 60 GB hard drive space is preferred"
```

install.ps1
```diff
@@ -479,102 +425,489 @@ if (-not $noGui.IsPresent) {
...
+		$EnoughHardStorageTooltip.AutoSize = $true
+		$EnoughHardStorageTooltip.location = New-Object System.Drawing.Point(15,$verticalPosition)
+		$EnoughHardStorageTooltip.Font = New-Object System.Drawing.Font('Microsoft Sans Serif',10)
+		$EnoughHardStorageTooltip.ForeColor = $grayedColor
+		$verticalPosition += 44
+
+		# $internetConnectivityTooltip
+		$internetConnectivityTooltip = New-Object system.Windows.Forms.Label
+		$internetConnectivityTooltip.text = "Valid Internet connectivity (mandatory)"
```
Ana06, line 699: [nitpick]
```suggestion
		$internetConnectivityTooltip.text = "You must have internet connection (mandatory)"
```

sara-rn:
Screenshot From 2025-08-16 13-15-55

Ana06 (approved):
Tested locally, everything looks good. Thanks for all the work @sara-rn to add the checks to the UI :rocket:

PR #713, opened 2025-08-04
[LayoutModification] Fix procmon and procexp

Ana06:
Fix procmon and procexp shortcuts in the taskbar.

### DIFF
--- a/LayoutModification.xml
+++ b/LayoutModification.xml
@@ -37,8 +37,8 @@
           <taskbar:DesktopApp DesktopApplicationLinkPath="%TOOL_LIST_DIR%\Disassemblers\ida.lnk"/>
           <taskbar:DesktopApp DesktopApplicationLinkPath="%TOOL_LIST_DIR%\Networking\fakenet.lnk"/>
           <taskbar:DesktopApp DesktopApplicationLinkPath="%TOOL_LIST_DIR%\PE\CFF Explorer.lnk"/>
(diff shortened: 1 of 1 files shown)


emtuls (changes requested):
This looks to be a correct change that we need to implement (I think previously we meant to point to 64 bit versions directly in `C:\Tools\sysinternals`).

But, I think we also need to make the current shortcuts of `"%TOOL_LIST_DIR%\Utilities\procexp.lnk"` and `"%TOOL_LIST_DIR%\Utilities\procmon.lnk"` point directly to the 64 bit versions in `C:\Tools\sysinternals`, otherwise we run into a previous issue where we have duplicate icons on the taskbar when the tool is running, which was https://github.com/mandiant/VM-Packages/issues/1100 

Essentially:
If you execute the 32 bit version of either tools at `C:\Tools\sysinternals\procmon`, it will in turn execute the 64 bit version when it detects that it's running on a 64 bit OS, which will cause the 32 bit (shortcut icon) stay on the taskbar and show a duplicate icon for the 64 bit version that is running.


**NOTE:** This then brings up a question of, how do we handle this for 32 bit installs if we're forcing our shortcut to point to the 64 bit version? I don't have a current 32 bit install to test if the 64 bit tool would be installed and if so, would the shortcut even work?)
Normally, pointing to the non-64 bit version which launches the 64 bit one anyway would be fine, but because we use shortcuts on the taskbar, it then becomes an issue. :(

emtuls (approved):

PR #714, opened 2025-08-04
Add projectUrl at the end of the description in the package selection

### LINKED ISSUE #708: [Installer]] Add tool link after the tool description
### Details

Add the tool link after the tool description as we do in the [_Packages_ Wiki] page(https://github.com/mandiant/VM-Packages/wiki/Packages) already.

sara-rn:
Add `projectUrl` at the end of the package description if it contains a valid URL to the package selection window
closes #708

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -520,25 +520,31 @@ if (-not $noGui.IsPresent) {
              $category = $_.SelectSingleNode("m:properties/d:Tags", $ns).InnerText
              # Select only packages that have the latest version, contain a category and the category is not excluded
              if (($isLatestVersion -eq "true") -and ($category -ne "") -and ($excludedCategories -notcontains $category)) {
-                  $packageName = $_.properties.Id
-                  $description = $_.properties.Description
+	             $packageName = $_.properties.Id
+				 $description = $_.properties.Description
+				 $projectUrl = $_.properties.projectUrl
 
-                  # Initialize category as an empty array
-                  if (-not ($packagesByCategory.ContainsKey($category))) {
-                     $packagesByCategory[$category] = @()
-                  }
-		  # Add the PackageName and PackageDesccription to each entry in the array
-                  $packagesByCategory[$category] += [PSCustomObject]@{
-                     PackageName = $packageName
-                     PackageDescription = $description
-                  }
-               }
-          }
+				 # Initialize category as an empty array
+				 if (-not ($packagesByCategory.ContainsKey($category))) {
+					 $packagesByCategory[$category] = @()
+				 }
+				 $packageObject = [PSCustomObject]@{
+				 PackageName        = $packageName
+				 PackageDescription = $description
+				 }
+				 # Check if $projectUrl contains a valid URL
+				 if ($projectUrl -match "^http") {
+					 Add-Member -InputObject $packageObject -MemberType NoteProperty -Name "PackageUrl" -Value $projectURl
+				 }
+				 # Add the PackageName and PackageDescription (and PackageUrl if present) to each entry in the array
+				 $packagesByCategory[$category] += $packageObject
+            }
+		  }
           # Check if there is a next link in the XML and set the API URL to that link if it exists
           $nextLink = $vm_packages.SelectSingleNode("//atom:link[@rel='next']/@href", $ns)
           $vmPackagesUrl = $nextLink."#text"
 
-       } while ($vmPackagesUrl)
+        } while ($vmPackagesUrl)
 
       return $packagesByCategory
     }
@@ -794,11 +800,6 @@ if (-not $noGui.IsPresent) {
         )
         return $packagesByCategory[$category]
     }
-    # Function that returns an array of all the packages that are displayed sorted by category from $packagesByCategory
-    function Get-AllPackages{
-        $listedPackages = $packagesByCategory.Values | ForEach-Object { $_ } | Select-Object -ExpandProperty PackageName
-        return $listedPackages
-    }
 
     # Function that returns additional packages from the config that are not displayed in the textboxes
     # which includes both Choco packages and packages from excluded categories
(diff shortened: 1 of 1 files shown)


sara-rn:
Space between package description and URL cannot be extended, I modified the Y position as per the comment in the issue #708
Screenshot From 2025-08-05 12-30-18

install.ps1
```diff
@@ -1003,6 +1012,23 @@ if (-not $noGui.IsPresent) {
...
+			$linkProjectUrl.Left = $checkbox.Right - 3
+			$linkProjectUrl.AutoSize                    = $true
+			$linkProjectUrl.Font                        = New-Object System.Drawing.Font('Microsoft Sans Serif',10)
+			$linkProjectUrl.LinkColor                   = "BLUE";
+			$linkProjectUrl.ActiveLinkColor             = "RED"
+			$linkProjectUrl.Text                        = $url
+			# Create a temporary, local variable with the correct URL for this specific link.
+			$linkProjectUrl.add_Click({
+				$tempUrl = $this.Text
+				[system.Diagnostics.Process]::start($tempUrl)
```
Ana06, line 1028: [nitpick] Why are you using `[system.Diagnostics.Process]::start` instead of the PS function`Start-Process`? This is already used in other links in the installer and I would change it there too.

Ana06 (changes requested):

install.ps1
```diff
@@ -520,19 +520,28 @@ if (-not $noGui.IsPresent) {
...
+		 if ($projectUrl -match "^http") {
+		     $packageObject = [PSCustomObject]@{
+		     PackageName        = $packageName
+		     PackageDescription = $description
+		     PackageUrl = $projectUrl}
+		 }else{
+		     $packageObject = [PSCustomObject]@{
+		     PackageName        = $packageName
+		     PackageDescription = $description}
+		 }
```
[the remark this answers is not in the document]
sara-rn: It's better to create `PSCustomObject` entries from the hashes for several reasons, including better readability and a more structured data format. Also the code is simpler and easier to read in the function `Get-AllPackages` (line 554) where yo need to get a flattened array of only entries with property `PackageName`.
Additionally this is what Gemini responds to the question: 

> Hash vs. PSCustomObject
> It's generally a better practice to use PSCustomObject entries instead of raw hashes for several reasons:
> 
> Property Access: You can access properties with the dot notation (e.g., $entry.PackageName), which is cleaner and more readable than hash notation ($entry["PackageName"]).
> 
> IntelliSense: In many PowerShell editors, PSCustomObject provides IntelliSense for its properties, which makes writing and navigating code much easier.
> 
> Object-Oriented Structure: They behave more like objects, which is often a more natural fit for representing structured data like your package entries.
> 
> Cmdlet Compatibility: Many PowerShell cmdlets are designed to work seamlessly with objects, making it easier to pipe and process your data.
> 
> While hashes are flexible, converting them to PSCustomObjects is a recommended practice for better code structure and compatibility.

sara-rn:
Thanks @Ana06 for the review, I addressed some of the requested changes, see my comments above. 
Screenshot From 2025-08-08 12-28-12

install.ps1
Ana06: @sara-rn what do you mean that it seems slower? We should use the fastest option if this is noticeable for the user.

sara-rn: I didn't experience any difference myself, this is what Gemini responded and maybe it's worth testing it

install.ps1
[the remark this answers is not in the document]
Ana06: The reason Gemini is providing are IMO not relevant for this simple code and not even correct. For example you can use the dot notation for a simple hash table too:
```
> $hash = @{ name = "value" }
> $hash.name
value
```

While I think AI is a useful resource that can speed your code writing, I advise you to be careful with trust it blindly and always rely on official documentation and testing the code on your own.

I think it is overkilled to use an object for this use of case. But if you think the code in `Get-AllPackages` is simpler, I have no strong opinion (note the suggestion of using a hash was a nitpick, only the code duplication was not, which is already solved.

install.ps1
```diff
@@ -520,25 +520,31 @@ if (-not $noGui.IsPresent) {
...
+				 if (-not ($packagesByCategory.ContainsKey($category))) {
+					 $packagesByCategory[$category] = @()
+				 }
+				 $packageObject = [PSCustomObject]@{
+				 PackageName        = $packageName
+				 PackageDescription = $description
+				 }
+				 # Check if $projectUrl contains a valid URL
+				 if ($projectUrl -match "^http") {
+					 Add-Member -InputObject $packageObject -MemberType NoteProperty -Name "PackageUrl" -Value $projectURl
```
Ana06, line 537: Why not just the code I proposed (simpler and more readable)?
```suggestion
					 $packageObject.PackageUrl = $projectUrl
```

install.ps1
Ana06: I  think performance shouldn't be an issue for such a simple task (opening a link) and as long as we are using it consistently I have no strong opinion (note I added it as nitpick), but please test that there is no performance difference locally or double check this information in official documentation.

install.ps1
sara-rn: This is because I am using the `PsCustomObject`, see my comment https://github.com/mandiant/flare-vm/pull/714#discussion_r2262522105
That doesn't work when using a `PsCustomObject`, see error : `ForEach-Object : Exception setting "PackageUrl": "The property 'PackageUrl' cannot be found on this object. Verify that the property exists and can be set."`

Ana06: As I said above, do not use `PsCustomObject` (instead of a simple hash table) if there is no reason to (I do not see a reason here), specially if it complicates the code, as it happens here. The code with a `PsCustomObject` is much more difficult to read than with a simple hash table.

install.ps1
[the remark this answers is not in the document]
sara-rn: @Ana06 I don't think using `PsCustomObject` is a overkill, please see the old installer making use of PS objects as well that served as a base for the new improved installer: https://github.com/mandiant/flare-vm/commit/de1dde81b7f9fba3a6a73d267ed02a4c6cb28d1f#diff-50d51a233d8ff4886061fc5cf1a03642de1f64b3625903e4f876e8cd2a3f5806L487
as per the documentation, `PsCustomObject` are the modern, more concise, and preferred way to create a custom object in PowerShell, while `New-Object -TypeName psobject` is the traditional, more verbose way to create a generic object in PowerShell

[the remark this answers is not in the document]
Ana06: > as per the documentation, PsCustomObject are the modern, more concise, and preferred way to create a custom object in PowerShell, while New-Object -TypeName psobject is the traditional, more verbose way to create a generic object in PowerShell

The way of creating the custom objects is ok. I just think that we do not need a custom object, but a simple hash. But I added this as nitpick in the first place anyway. Let me test the current code locally and if it works as expected we can merge the PR.

PR #721, opened 2025-08-08
[CI] vboxcommon: Clean environment for subprocess

Ana06:
When running as a PyInstaller bundle, the bootloader sets the `LD_LIBRARY_PATH` environment variable to point to its temporary directory. This can force `VBoxManage` to use the wrong libraries, causing it to fail.

This change creates a clean environment for the `VBoxManage` subprocess. It restores the original `LD_LIBRARY_PATH` (if available) or removes it, ensuring `VBoxManage` can find and use the correct system libraries.


stevemk14ebr (approved):

stevemk14ebr:
Might also want to remove the pgi import install if it's not necessary

stevemk14ebr (approved):

Issue #737, opened 2025-12-10
Flare-VM Package Installation Windows UI Issue


PrajeetGuha:
### What's the problem?

The Flare VM Package Installation Window that pop-ups during installation is vertically too long to fit in smaller display PCs (sometimes an issue in VM been spawned). While the window been big is not the real issue - scrolling to the bottom to click "Install" button or accessing any of the options at the bottom is not possible unless vertical height is not long enough.
I use a display VM with 1280x720 and the window does not fit in to click the "Install" button and doing the window resizing does not work. The scroll bar for the window is only in the package selection portion.

<img width="1306" height="727" alt="Image" src="https://github.com/user-attachments/assets/cb190ff9-73ae-45af-a08c-f6b3a7ebcaf2" />

### Steps to Reproduce

Occurs during Flare VM installation

### Environment

- Virtualization Software: VMWare Workstation 17
- VM OS version: Windows 11 Pro

### Additional Information

The workaround for the issue is to change the resolution to fit in the UI.
Three possible solutions (not verified) to it:
- Add a scrollbar to the component outside the existing scrollbar for package selection to allow scrolling the main component.
- Move the lower configuration settings to the next window in the wizard.
- Make the UI components resolution dependent so that the package selection component resize based on resolution of the machine in which it is running.

divyesh-git22:
Hi, I’d like to work on this issue. My proposed fix:

Make the installer window responsive to screen
Add a scrollbar for the entire UI so the bottom buttons remain usable

If everything looks okay, I can start working on this issue

emtuls:
Thanks for the bug @PrajeetGuha! That is certainly frustrating. I will try to take a look at that in due time, though other priorities may come first.

In the mean time, while not ideal, there is a `-noGUI`  flag that you can use that will install the default tools listed in the [config.xml
](https://github.com/mandiant/flare-vm/blob/main/config.xml)

If you require additional tools beyond the default, you can view the full list (with their descriptions) [here](https://github.com/mandiant/VM-Packages/wiki/Packages).

Once you know the additional tools you would like to add, you can make a copy of the [config.xml
](https://github.com/mandiant/flare-vm/blob/main/config.xml) and amend those tools to the `<packages>` section, then pass that in as an argument to the install script with `-customConfig`: https://github.com/mandiant/flare-vm/tree/main?tab=readme-ov-file#installer-parameters

emtuls:
[...]

PR #740, opened 2025-12-26
Fix installer pre-check UI layout by using scrollable panel

divyesh-git22:
This PR refactors the Pre-Install Checks UI to use a scrollable Panel
instead of fixed positioning.
what this change does:
- Prevents UI overflow on smaller screens
- Removes dependency on manual spacing
No installer logic was changed by myside

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -433,190 +433,48 @@ if (-not $noGui.IsPresent) {
 		################################ Installer Checks Form Controls #################################
 		#################################################################################################
 
-		$formChecksManager           = New-Object system.Windows.Forms.Form
-		$formChecksManager.ClientSize  = New-Object System.Drawing.Point(700,640)
-		$formChecksManager.text      = "FLAREVM Pre-Install Checks"
-		$formChecksManager.TopMost   = $true
-		$formChecksManager.StartPosition = 'CenterScreen'
+		$formChecksManager = New-Object System.Windows.Forms.Form
+$formChecksManager.ClientSize = New-Object System.Drawing.Point(700,640)
+$formChecksManager.Text = "FLAREVM Pre-Install Checks"
+$formChecksManager.StartPosition = 'CenterScreen'
 
-		$ChecksPanel                     = New-Object system.Windows.Forms.Panel
-		$ChecksPanel.height              = 460
-		$ChecksPanel.width               = 89
-		$ChecksPanel.location            = New-Object System.Drawing.Point(570,8)
+$ChecksPanel = New-Object System.Windows.Forms.Panel
+$ChecksPanel.Dock = 'Fill'
+$ChecksPanel.AutoScroll = $true
 
-		$InstallChecksGroup              = New-Object system.Windows.Forms.Groupbox
-		$InstallChecksGroup.height       = 490
-		$InstallChecksGroup.width        = 665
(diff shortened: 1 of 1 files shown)


emtuls:
Thank you for the PR @divyesh-git22! 

I am testing the PR on Windows 10 and I did notice some issues that need to be addressed:

1. **The `Pre-Install Checks` window seems to be broken:**
Scroll wheels have been added to the sides, which is a great change! But unfortunately, some other things broke in the process:
- The checkbox we had to ensure users have made a snapshot before continuing seems to be missing.
- The `Continue` and `Cancel` boxes seem to be missing.
- The `True` and `False` indicators next to each check also seem to be missing.

Your code:
image
image

This is how I would expect it to look when fixed:
image

2. **The internet connectivity checks seem to have been removed:**

Before (original code):
image

After (your PR):
image


3. **The package selection window still doesn't include the scroll bars on the sides, which I believe was the main concern from #737** 

Your code (note there are no scroll wheels on the sides for lower resolutions):
image

This is how I would expect it to look when fixed:
image

emtuls (changes requested):
See previous comment!

PrajeetGuha:
@divyesh-git22 I checked your code. Looks like your edits are on pre-install check window and not on the installation wizard that has the package selection, installation path, etc. configurations.
My issue is with the **VM package selection** window that overflows and prevents me to click the OK button at the bottom as stated by @emtuls .

divyesh-git22:
Hii @emtuls and @PrajeetGuha,

I sincerely apologize for the confusion and the broken UI in my previous commits. As I am new to the project, I misunderstood which window needed the scrollbar and accidentally broke the Pre-Install UI and internet check logic.

I am currently reverting install.ps1 to its original state to fix these errors. Once the project is back to a working state, I will correctly apply the scrollable fix to the Package Selection window as originally requested in #737.

Thank you for your patience and for the detailed feedback

emtuls:
Hi @divyesh-git22!

I see that you are making a bunch of commits which might be unnecessary. 🙂 Also, we prefer to only have 1 commit made per change (i.e., 1 commit for this PR) to keep a cleaner git history.

Might I suggest one of the following to clean up the branch?

**Option 1: Start fresh (Deletes your current changes).** 
Use this to wipe the slate and re-apply your changes cleanly:
- `git reset --hard main`
- make your changes to the file
- `git add <file>`
- `git commit -m "Clean, single commit message"`
- `git push origin <your-branch-name> --force-with-lease`


**Option 2: Squash and Reset (Keeps your work, but cleans history)**
Alternatively, this would revert you back to the main commit, keeping your changes, so that you can make a single commit:
- `git reset --soft main`
- `git commit -m "Clean, single commit message"`
- `git push origin <your-branch-name> --force-with-lease`

divyesh-git22:
Hii @emtuls,

Thank you so much for your guidance and for helping me ! I sincerely apologize for the earlier confusion and any extra work it caused you, I appreciate your patience while I was learning the process.

I have now followed Option 2 and squashed all changes into a single, clean commit. I also moved the fix to the correct Package Selection window as requested:

Added a scrollbar to the categories panel.

Adjusted the window size and moved the buttons (Install/Cancel) up to ensure they are visible on all screen resolutions.
The PR should be much cleaner and ready for your review now. Thank you again for all your help

PR #747, opened 2026-04-28
[virtualbox] Support IDA >=9.2 in REMnux

Ana06:
Install `libxcb-cursor0` in REMnux, as otherwise IDA >=9.2 fails with:
```
remnux@remnux:~$ /home/remnux/ida-pro-9.3/ida

qt.qpa.plugin: From 6.5.0, xcb-cursor0 or libxcb-cursor0 is needed to load the Qt xcb platform plugin.

qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.

This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
```

### DIFF
--- a/virtualbox/configs/remnux.yaml
+++ b/virtualbox/configs/remnux.yaml
@@ -6,7 +6,7 @@ SNAPSHOT:
 CMDS:
   - |
     # Install additional useful packages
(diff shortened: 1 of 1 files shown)


emtuls (approved):
Awesome catch. Thanks @Ana06 ! :)

